PR #1, opened 2015-06-23
remove bin, modify test

vaderyang:
Just a first commit for participating this project. Hope I can do more.

1. The Linux binary I think it is not necessary to be included in the repository, so I removed it.
2. The input file test cannot be compiled as there is a mismatch for parameter type of Type() which is currently not a function. Simply alter the Type() to Type will solve this issue and have the test passed.

### DIFF
--- a/input/file/inputfile_test.go
+++ b/input/file/inputfile_test.go
@@ -27,7 +27,7 @@ func Test_main(t *testing.T) {
 	if len(inputs) > 0 {
 		input := inputs[0].(*InputConfig)
 		assert.IsType(&InputConfig{}, input)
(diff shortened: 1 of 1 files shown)


tsaikd:
Thank for the PR.

1. the binary is used for building docker image automatically at docker hub, so it's necessary. now I create a new branch "binary" for it, and reset master to HEAD~1. you can fetch again for the latest master branch.
2. I review all test files and fix them at the new commit, thanks again.

vaderyang:
Thanks for pointing out the docker issue, I didn't realize this.  Appreciate for the prompt response.

PR #5, opened 2016-04-19
AMQP output

mirdhyn:
First attempt for a AMQP output plugin.

### DIFF
--- a/modloader/modloader.go
+++ b/modloader/modloader.go
@@ -7,6 +7,7 @@ import (
 	"github.com/tsaikd/gogstash/input/exec"
 	"github.com/tsaikd/gogstash/input/file"
 	"github.com/tsaikd/gogstash/input/http"
+	"github.com/tsaikd/gogstash/output/amqp"
 	"github.com/tsaikd/gogstash/output/elastic"
 	"github.com/tsaikd/gogstash/output/redis"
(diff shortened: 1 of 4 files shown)


config/output.go
```diff
@@ -36,7 +36,7 @@ func (t *Config) RunOutputs() (err error) {
 func (t *Config) runOutputs(evchan chan logevent.LogEvent, logger *logrus.Logger) (err error) {
 	outputs, err := t.getOutputs()
 	if err != nil {
-		return errutil.New("get outputs failed", err)
+		return err
```
tsaikd: I think it's better to wrap the output module error, because the output module may not throw enough error information.

mirdhyn: I would agree but when running the test - without rabbitmq, elasticsearch nor redis - I saw that `errutil.New("get outputs failed", err)` would only carry `"get outputs failed"`.

See gist for the version with only `return err`: 
https://gist.github.com/mirdhyn/2bea74ddfeaf689aa6c1a1ce4ef626a8
and the one with errutil:
https://gist.github.com/mirdhyn/c519636776e4224334f1c8d7af87ed69

mirdhyn: Anyway, let's check that another time... I'll make this PR only about the AMQP output.

PR #12, opened 2016-05-06
Filters!

mirdhyn:
Big PR, sorry... but filters are really needed!


tsaikd:
Good PR!
But I think it's better to use `filter/addfield/filteraddfield` instead of `filter/add_field/filteraddfield` for consistent of module naming.

mirdhyn:
sure, I updated the PR.

PR #22, opened 2016-11-21
add output=email plugin and fix input=file bug

clevertension:
add output=email plugin and fix input=file bug


tsaikd:
Could you send only one purpose for each PR? I think this one contains 3 irrelevant commits.
1. change inputfile module
2. add outputemail module
3. fix stdout README typo

clevertension:
OK, i will give another PR

PR #27, opened 2017-01-12
Add redis input, http listen input, ES5.0 output support

thecubed:
Documentation/tests to come soon. Just adding this as a PR for now so it's visible in case anyone wants to use it in the meantime.
Feel free to not merge this until the documentation, examples, and tests are ready.

Oh, and I added a `--debug` flag to the app, maybe that wasn't necessary, but I couldn't figure out how debug logs were supposed to be enabled previously.

The Redis input allows you to use Gogstash in a setup similar to this:
- Application does a RPUSH to redis for log statements (an example for NodeJS is to use Winston-Redis as an output). 
This could be instead another Gogstash instance using an HTTP input (included in this changeset) that is writing to redis.
- Gogstash reads from Redis using the redis input and outputs to Elasticsearch

The HTTP input allows you to POST data to Gogstash, and it has a few configurable settings (such as the path you want it listening on, and any headers you might require for auth).
I have not yet added support for the HTTP input to have multiple paths on the same port (i.e. if you listen on port 8081 with path /, you can't have a path /log on port 8081).

In my case, I'm using a Gogstash instance on an edge server with the HTTP input to write to a redis list and a Gogstash instance (on an internal server) to pick up data from the list and write it to Elasticsearch. This allows me to scale my worker instances for resiliency/security.

ES5.x support has been added - it'll autodetect your ES version if possible.

Thanks for Gogstash!

### DIFF
--- a/output/elastic/outputelastic.go
+++ b/output/elastic/outputelastic.go
@@ -3,7 +3,14 @@ package outputelastic
 import (
 	"github.com/tsaikd/gogstash/config"
 	"github.com/tsaikd/gogstash/config/logevent"
-	"gopkg.in/olivere/elastic.v3"
+	elastic3 "gopkg.in/olivere/elastic.v3"
+	elastic5 "gopkg.in/olivere/elastic.v5"
+	"net/http"
+	"github.com/Sirupsen/logrus"
+	"io/ioutil"
+	"encoding/json"
+	"github.com/hashicorp/go-version"
+	"golang.org/x/net/context"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


output/elastic/outputelastic.go
```diff
@@ -29,19 +38,40 @@ func DefaultOutputConfig() OutputConfig {
...
-		elastic.SetSniff(conf.Sniff),
-	)
+	switch (conf.ElasticVersion) {
+	case "auto":
+		conf.clientVersion = detectElasticVersion(&conf, logger)
+	case "3":
+		conf.clientVersion = 3
+	case "5":
+		conf.clientVersion = 5
+	default:
```
thecubed: 'Auto' is indeed used as the default per DefaultOutputConfig

output/elastic/outputelastic.go
```diff
@@ -50,16 +80,57 @@ func InitHandler(confraw *config.ConfigRaw) (retconf config.TypeOutputConfig, er
...
 	doctype := event.Format(t.DocumentType)
 	id := event.Format(t.DocumentID)
 
-	_, err = t.client.Index().
-		Index(index).
-		Type(doctype).
-		Id(id).
-		BodyJson(event).
-		Do()
+	if t.clientVersion == 3 {
```
thecubed: I can do that if you'd prefer (or if it's better Golang convention).

tsaikd:
Could you use `go fmt` tool for the source code?
See [CI log](https://travis-ci.org/tsaikd/gogstash/jobs/191456363) for more information

Issue #28, opened 2017-03-21
codec support


jetgeng:
Is there any support for codec plugin? Or is there any plan?

tsaikd:
Which codec do you need? It's better to say the whole story.

jetgeng:
There is a [codec!](https://www.elastic.co/guide/en/logstash/current/codec-plugins.html) system  in logstash. We can use codec to  changes the data representation, and we can put codec in input, filter and output. I think it's very useful, and i did't find this in gogstash. So i hope gogstash can support it!

tsaikd:
It's hard to implement all features in logstash, because it's a side project for me. But if you want, PR is welcomed. I didn't use codec in logstash. My use case is very simple: read nginx log and write to elastic, etc. You can show your use case for discussing.

aantono:
I’ll try to give it a go to add base support for codecs, but doubtfully will implement all the ones Logstash offers.

aantono:
@jetgeng I'm working on adding the foundation for Codecs, do you have any particular codecs in mind that you wanted to see be implemented?

jbuchbinder:
The multiline codec would be useful to properly implement multiple line Java stack trace parsing.

PR #35, opened 2018-02-06
Add initial foundation for Codec support.

aantono:
This will build a foundation for Codec support asked in #28

### DIFF
--- a/config/codec.go
+++ b/config/codec.go
@@ -43,7 +43,10 @@ func RegistCodecHandler(name string, handler CodecHandler) {
 
 // GetCodec returns a codec based on the 'codec' configuration from provided 'ConfigRaw' input
 func GetCodec(ctx context.Context, raw ConfigRaw) (TypeCodecConfig, error) {
-	codecConfig, _ := dyno.Get(map[string]interface{}(raw), "codec")
+	codecConfig, err := dyno.Get(map[string]interface{}(raw), "codec")
+	if err != nil {
+		Logger.Infof("No codec has been configured for input %v", raw)
+	}
 
 	if codecConfig == nil {
 		return nil, nil
--- a/config/codec.go
+++ b/config/codec.go
@@ -1,0 +1,102 @@
+package config
+
+import (
+	"context"
+	"time"
+
+	"github.com/icza/dyno"
+	"github.com/tsaikd/KDGoLib/errutil"
+	"github.com/tsaikd/gogstash/config/logevent"
+)
+
+// errors
+var (
+	ErrorUnknownCodecType1 = errutil.NewFactory("unknown codec config type: %q")
+	ErrorInitCodecFailed1  = errutil.NewFactory("initialize codec module failed: %v")
+)
+
+// TypeCodecConfig is interface of codec module
+type TypeCodecConfig interface {
+	TypeCommonConfig
+	// The codec’s decode method is where data coming in from an input is transformed into an event.
+	Decode(ctx context.Context, data []byte, extra map[string]interface{}, msgChan chan<- logevent.LogEvent) (ok bool, err error)
+	// The encode method takes an event and serializes it (encodes) into another format.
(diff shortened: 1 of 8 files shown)


config/codec.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+// RegistCodecHandler regist a codec handler
+func RegistCodecHandler(name string, handler CodecHandler) {
+	mapCodecHandler[name] = handler
+}
+
+// GetCodec returns a codec based on the 'codec' configuration from provided 'ConfigRaw' input
+func GetCodec(ctx context.Context, raw ConfigRaw) (TypeCodecConfig, error) {
+	codecConfig, _ := dyno.Get(map[string]interface{}(raw), "codec")
```
tsaikd: Why not catch error?
> codecConfig, `err` := dyno.Get(map[string]interface{}(raw), "codec")

input/file/inputfile.go
```diff
@@ -69,6 +69,11 @@ func InitHandler(ctx context.Context, raw *config.ConfigRaw) (config.TypeInputCo
 		return nil, err
 	}
 
+	conf.Codec, err = config.GetCodec(ctx, *raw)
```
tsaikd: forget to handle error?
```
if err != nil {
    return nil, err
}
```

config/codec.go
aantono: There is no way that the error will occur, as we know for a fact that we are passing in a `map[string]interface{}`, so internally all it will do is to remap the structure to a nested `map[string]interface{}` instead of `map[interface]interface{}`  But I would be happy to add the check if needed.

input/file/inputfile.go
aantono: Here we don't actually care about the error, because if no codec is configured, the internal mapping will cause an error, but it is not relevant (IMHO).  If there is no Codec, for one reason, or another, we just use the default, no?

input/file/inputfile.go
```diff
@@ -69,6 +69,11 @@ func InitHandler(ctx context.Context, raw *config.ConfigRaw) (config.TypeInputCo
...
+	if conf.Codec == nil {
+		conf.Codec, _ = config.DefaultCodecInitHandler(nil, nil)
```
tsaikd: Please handle error everywhere even you know it will not return error when writing golang.
Because other developers may change the function in the future.

input/file/inputfile.go
```diff
@@ -217,22 +222,33 @@ func (t *InputConfig) fileReadLoop(
...
-			Extra: map[string]interface{}{
+		//event := logevent.LogEvent{
+		//	Timestamp: time.Now(),
+		//	Message:   line,
+		//	Extra: map[string]interface{}{
+		//		"host":   t.hostname,
+		//		"path":   fpath,
+		//		"offset": since.Offset,
+		//	},
+		//}
```
tsaikd: If the code is not necessary, how about remove it?

input/file/inputfile.go
```diff
@@ -217,22 +222,33 @@ func (t *InputConfig) fileReadLoop(
...
+			map[string]interface{}{
 				"host":   t.hostname,
 				"path":   fpath,
 				"offset": since.Offset,
 			},
-		}
+			msgChan)
 
-		since.Offset += int64(size)
+		if err == nil {
```
tsaikd: Please handle error, maybe just a logger.

config/codec.go
aantono: So this line definitely does not need error checking, because the `codec` config is optional, and if it is not present, the `dyno.Get` call will return an error, and we don't want to throw it... I can log it as an `Info`, but that's about it.  Want me to do that?

aantono: The default behavior is delegated to the specific `input` implementation... Not everything can/will have a default codec, so I didn't want to ALWAYS return a codec, but rather return `nil` if provided `ConfigRaw` does not have a `codec` key.

PR #37, opened 2018-02-08
Add MongoDB output module

mickar:
Add MongoDB output module

### DIFF
--- a/output/mongodb/outputmongo.go
+++ b/output/mongodb/outputmongo.go
@@ -1,0 +1,157 @@
+package outputmongodb
+
+import (
+	"context"
+	"time"
+	"sync"
+
+	"github.com/tsaikd/KDGoLib/timeutil"
+	"github.com/tsaikd/KDGoLib/errutil"
+	"github.com/tsaikd/gogstash/config"
+	"github.com/tsaikd/gogstash/config/logevent"
+	"gopkg.in/mgo.v2"
+)
+
+// ModuleName is the name used in config file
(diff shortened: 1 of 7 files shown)


tsaikd:
Please use gofmt or other tools like goimports to normalize source codes.

output/mongodb/outputmongo.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		session := t.pool.Get()
+		coll := session.DB(t.Database).C(t.Collection)
+		err := coll.Insert(m)
+		if err == nil {
+			t.pool.Put(session)
+			return nil
+		}
+		ErrorInsertMongoDBFailed1.New(err, event)
```
tsaikd: What do you mean for this line? Create an error without return it?

mickar: At the beginning, I wrote another code and I returned this error. When I saw the redis output module, I modified my code to be "more compliant" with a similar module. I actually forgot to remove this error :x I'm sorry.
```
func (t *OutputConfig) Output(ctx context.Context, event logevent.LogEvent) error {
	m := event.GetMap()
	session := t.pool.Get()
	coll := session.DB(t.Database).C(t.Collection)
	err := coll.Insert(m)
	if err == nil {
		t.pool.Put(session)
		return nil
	}
	t.pool.Put(session)
	return ErrorInsertMongoDBFailed1.New(err, event)
}
```
What do you think about the code above ? Is it necessary to try more than once ? I don't know.

PR #40, opened 2018-07-17
add filter/cond and output/cond

tengattack:
Hi, I think this could help us to do more complex consumption for events.

Add `filter/cond` and `output/cond` implements the sequence control for event's filters and outputs.

a simple config:
```yml
input:
  - type: httplisten
    address: '0.0.0.0:8081'
    path: '/log'
filter:
  - type: add_field
    key: instance_id
    value: '123456'
  - type: cond
    condition: "[nginx.access.url] != nil ? [nginx.access.url] =~ '^/api/'"
    filter:
      - type: add_field
        key: controller
        value: api
output:
  - type: cond
    condition: "info == 'ERROR'"
    output:
      - type: stdout
```

### DIFF
--- a/filter/cond/README.md
+++ b/filter/cond/README.md
@@ -1,0 +1,34 @@
+gogstash cond filter module
+=============================
+
+Condition syntax depends `govaluate`:
+
+* https://github.com/Knetic/govaluate
+
+Only boolean `true` for the result value is considered to be eligible.
+
+## Synopsis
+
+```yaml
+filter:
+  - type: cond
+    # (required) condition need to be satisfied
+    condition: "level == 'ERROR'"
+    # (required) filter config
+    filter:
+      - type: add_field
+        key: foo
+        value: bar
+```
+
+```yaml
+filter:
+  - type: cond
+    # (required) condition need to be satisfied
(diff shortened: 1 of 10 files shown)


tsaikd:
thx for the PR

filter/cond/README.md
```diff
@@ -0,0 +1,34 @@
...
+      - type: add_field
+        key: foo
+        value: bar
+```
+
+```yaml
+filter:
+  - type: cond
+    # (required) condition need to be satisfied
+    condition: "[nginx.access.url] ? [nginx.access.url] =~ '^/api/'"
```
tengattack: Sorry the the typo in README, it should be != nil before ?
Just like the PR description

tengattack:
Could you help me to fix that?

filter/cond/README.md
tengattack: Sorry the the typo in README, it should no ternary check:
```yml
condition: "[nginx.access.url] =~ '^/api/'"
```
Just like those which in the PR description
Since the ternary operator of [govaluate](https://github.com/Knetic/govaluate) only accept `true` on left side:
https://github.com/Knetic/govaluate/blob/master/MANUAL.md#ternary-true-

PR #45, opened 2018-09-17
Fix event add tag not work when it has tags in extra

tengattack:
Hello,

I found that gogstash's `addTag` will not work when an event already has `tags` field, since it has a separated `Tags` property.

### DIFF
--- a/config/logevent/logevent.go
+++ b/config/logevent/logevent.go
@@ -42,11 +42,41 @@ func (t LogEvent) getJSONMap() map[string]interface{} {
 	if t.Message != "" {
 		event["message"] = t.Message
 	}
-	if len(t.Tags) > 0 {
-		event["tags"] = t.Tags
-	}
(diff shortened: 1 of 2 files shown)


config/logevent/logevent.go
```diff
@@ -42,12 +42,42 @@ func (t LogEvent) getJSONMap() map[string]interface{} {
...
+						stringTags = append(stringTags, tag)
+					default:
+						ok = false
+						break tags_loop
+					}
+				}
+				if ok {
+					event["tags"] = append(stringTags, t.Tags...)
+				} else {
+					// TODO: warning
```
tsaikd: how about using `config.Logger.Warning()` here?

tsaikd: sorry, import cycle...logger should move to another package

tengattack: yep, we can make another commit to do that

PR #59, opened 2018-09-26
Add worker management

tengattack:
Since we create multiple worker processes, we need to monitor their state.

The main change is the main process (if we set multiple workers) will not process events any more but management of worker processes.

### DIFF
--- a/cmd/worker.go
+++ b/cmd/worker.go
@@ -12,7 +12,10 @@ func waitSignals(ctx context.Context) error {
 	osSignalChan := make(chan os.Signal, 1)
 	signal.Notify(osSignalChan, os.Interrupt, os.Kill)
 
-	sig := <-osSignalChan
-	goglog.Logger.Info(sig)
+	select {
+	case <-ctx.Done():
+	case sig := <-osSignalChan:
+		goglog.Logger.Info(sig)
+	}
 	return nil
 }
--- a/cmd/worker.go
+++ b/cmd/worker.go
@@ -1,0 +1,18 @@
+package cmd
+
+import (
+	"context"
+	"os"
+	"os/signal"
+
+	"github.com/tsaikd/gogstash/config/goglog"
+)
+
+func waitSignals(ctx context.Context) error {
+	osSignalChan := make(chan os.Signal, 1)
+	signal.Notify(osSignalChan, os.Interrupt, os.Kill)
+
+	sig := <-osSignalChan
+	goglog.Logger.Info(sig)
+	return nil
+}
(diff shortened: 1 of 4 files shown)


cmd/worker_unix.go
```diff
@@ -3,15 +3,71 @@
...
+	pids := make([]int, workers)
+	for i := 0; i < workers; i++ {
+		pid, err := startWorker(args, attr)
+		if err != nil {
+			return err
+		}
+		pids[i] = pid
+	}
+
+	go waitWorkers(ctx, pids, args, attr)
```
tsaikd: it's safer to use `errgroup.WithContext()` to create goroutine

tengattack: Do you mean like this:
https://github.com/tsaikd/gogstash/pull/59/commits/0f5f745ab1c2eeb13ab5c9dfa03826af60131065

tsaikd: Yes

cmd/worker.go
```diff
@@ -0,0 +1,18 @@
...
+	"os/signal"
+
+	"github.com/tsaikd/gogstash/config/goglog"
+)
+
+func waitSignals(ctx context.Context) error {
+	osSignalChan := make(chan os.Signal, 1)
+	signal.Notify(osSignalChan, os.Interrupt, os.Kill)
+
+	sig := <-osSignalChan
```
tsaikd: will app hang on this when ctx cancel ?

```
select {
case <- ctx.Done():
    return nil
case sig := <-osSignalChan:
    // blahblah...
```

tengattack: Oops, https://github.com/tsaikd/gogstash/pull/59/commits/e7314ff384035e4980250c22fd0a4231ae33c203 this will end with `cancel`, I think it will be ok?

tengattack: Sorry, I misunderstand, https://github.com/tsaikd/gogstash/pull/59/commits/18da6cd399e664387d5ae7c50d470719d80fecd6 this will be fine

tengattack: Oh! you are right, app will final do `cancel` this time https://github.com/tsaikd/gogstash/pull/59/commits/ee162c67dcdfc2c69ee4cb533a80ae023c3b8675

cmd/worker_unix.go
```diff
@@ -3,15 +3,74 @@
...
+	return
+}
+
+func waitWorkers(ctx context.Context, pids []int, args []string, attr *syscall.ProcAttr) error {
+	var ws syscall.WaitStatus
+	for {
+		// wait for any child process
+		pid, err := syscall.Wait4(-1, &ws, 0, nil)
+		if err != nil {
+			goglog.Logger.Error("wait4() error: %v", err)
```
tsaikd: do you mean `Errorf` ?

tengattack: Oh

PR #60, opened 2018-09-27
Add input beats

tengattack:
finally we can replace logstash :)


tsaikd:
I didn't use logstash for a few years...^_^

PR #64, opened 2018-10-16
grok: Add ability to accept multiple patterns for filter.

jbuchbinder:
Also breaks backwards configuration compatibility, but allows multiple patterns to be checked before an error is thrown by the grok filter.

### DIFF
--- a/filter/grok/filtergrok.go
+++ b/filter/grok/filtergrok.go
@@ -69,10 +69,11 @@ func (f *FilterConfig) Event(ctx context.Context, event logevent.LogEvent) logev
 		values, err := f.grk.Parse(thisMatch, message)
 		if err != nil {
 			continue
-		}
-
-		for key, value := range values {
-			event.SetValue(key, event.Format(value))
+		} else {
+			for key, value := range values {
+				event.SetValue(key, event.Format(value))
+			}
+			break
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


filter/grok/filtergrok.go
```diff
@@ -64,16 +64,24 @@ func InitHandler(ctx context.Context, raw *config.ConfigRaw) (config.TypeFilterC
 // Event the main filter event
 func (f *FilterConfig) Event(ctx context.Context, event logevent.LogEvent) logevent.LogEvent {
 	message := event.GetString(f.Source)
-	values, err := f.grk.Parse(f.Match, message)
+	var err error
+	for _, thisMatch := range f.Match {
+		values, err := f.grk.Parse(thisMatch, message)
```
tsaikd: 1. err may be shadow, so line 80 will not cache any error
2. err may be overwrite at the next for loop, so some error may be ignore

tsaikd:
this PR is not merged, but github say merged
if you still want to have this patch, please resend a PR

filter/grok/filtergrok.go
jbuchbinder: 1. I'm not sure whether the lexer would figure out that err already exist or if it needs values to be defined in a line beforehand. Stupid lexical ambiguity ...
2. The overwriting thing was intentional -- the idea is that we skip to the next pattern if we don't find one, otherwise we'd be writing a series of errors every time the first pattern wasn't matched but the fifth or sixth one was.

tsaikd: OK
But github think this PR was merged.
You should create a new PR for merging to master branch.

PR #66, opened 2018-10-18
filter/date: Improvements

jbuchbinder:
Add'l parameters to add more logstash-like compatibility.

### DIFF
--- a/filter/date/filterdate.go
+++ b/filter/date/filterdate.go
@@ -20,9 +20,9 @@ const ErrorTag = "gogstash_filter_date_error"
 type FilterConfig struct {
 	config.FilterConfig
 
-	Format string `json:"format"` // date parse format
-	Source string `json:"source"` // source message field name
-	Joda   bool   `json:"joda"`   // whether using joda time format
+	Format []string `json:"format"` // date parse format
+	Source string   `json:"source"` // source message field name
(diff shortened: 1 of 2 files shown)


filter/date/filterdate.go
```diff
@@ -35,8 +34,10 @@ func DefaultFilterConfig() FilterConfig {
 				Type: ModuleName,
 			},
 		},
-		Format: []string{time.RFC3339Nano},
-		Source: "message",
+		Format:       []string{time.RFC3339Nano},
+		Source:       "message",
+		Target:       "@timestamp",
+		TagOnFailure: []string{"gogstash_filter_date_error", "_dateparsefailure"},
```
tsaikd: Why TagOnFailure "_dateparsefailure" as a default value?
I mean there was "gogstash_filter_date_error"

jbuchbinder: That was specifically to ensure some sort of compatibility with the output that logstash's date plugin dumps into elasticsearch, so I had added it to the list of tags added on failure to not break compatibility with anything anyone had been doing.

tsaikd: I think it's better to remove from default value, because not all users migrate from logstash, and it's unnecessary for pure gogstash user.
Write some examples or hints in README for your case?

PR #67, opened 2018-10-18
output/elastic: Elasticsearch output improvements

jbuchbinder:


tengattack:
the readme should be updated as grok doesn't support multiple patterns right now:
https://github.com/tsaikd/gogstash/blob/master/filter/grok/filtergrok.go#L23

PR #74, opened 2019-01-25
Add event config

tengattack:
We can configure event to json action now~


tengattack:
Ummm, it seems output prometheus module failed again

PR #82, opened 2019-04-14
Allow bypass elasticsearch certificate validation

horacimacias:


horacimacias:
Sorry, perhaps I messed up as I was trying to create 2 pull requests:
1) allow ssl certificate bypass
2) have elastic URLs resolved against environment variables

PR #84, opened 2019-04-14
Resolve environment variables on elastic URLs

horacimacias:


horacimacias:
This allows configuration such as:

> url: ["https://%{LOGSTASH_ES_USERNAME}:%{LOGSTASH_ES_PASSWORD}@elastic-service:9200"]

Perhaps this "resolve environment properties in config values" should be done more generically, for all configs.

tsaikd:
Can you add an unit test?

output/elastic/outputelastic.go
```diff
@@ -105,6 +105,14 @@ func InitHandler(ctx context.Context, raw *config.ConfigRaw) (config.TypeOutputC
 	// map Printf to error level
 	logger := &errorLogger{logger: goglog.Logger}
 
+	// replace env var names with values on URL config
+	var resolvedUrls []string
+	for _, url := range conf.URL {
+		newUrl := logevent.FormatWithEnv(url)
+		resolvedUrls = append(resolvedUrls, newUrl)
+	}
+	conf.URL = resolvedUrls
```
tengattack: Maybe we should not change the original config url string.
BTW, you should use URL instead of Url by following go lint & goreturns rules.

output/elastic/outputelastic.go
```diff
@@ -105,8 +105,15 @@ func InitHandler(ctx context.Context, raw *config.ConfigRaw) (config.TypeOutputC
 	// map Printf to error level
 	logger := &errorLogger{logger: goglog.Logger}
 
+	// replace env var names with values on URL config
+	var resolvedURLs []string
+	for _, URL := range conf.URL {
```
tengattack: Sorry about my words, we should uppercase URL only if the variable name not starts with it, the variable `URL` should be `url` here.

horacimacias: no worries. done

output/elastic/outputelastic_test.go
```diff
@@ -20,6 +24,35 @@ func init() {
...
+  - type: elastic
+    url: ["%{MYVAR}"]
+    index: "gogstash-index-test"
+    document_type: "testtype"
+    document_id: "%{fieldstring}"
+    bulk_actions: 0
+	`)))
+	a.Nil(err)
+	a.NotNil(conf)
+	_, err = InitHandler(context.TODO(), &conf.OutputRaw[0])
```
tengattack: Could you check the config will resolve to `ts.URL` successfully after here?

horacimacias: I'm not sure.
Do you mean something like this?

```suggestion
c, err := InitHandler(context.TODO(), &conf.OutputRaw[0])
a.Nil(err)
outputConfig := c.(*OutputConfig)
a.Equal(ts.URL, outputConfig.URL)
```
the problem is I'm not changing the original url config string as you suggested, so not sure how to actually check it. Elastic client doesn't have anything that I can see allowing access of the url either.

tengattack: Oh, sorry, It will not save resolved url to config now, I made a mistake.

tengattack: But maybe we could create a function for resolving urls and test that function? Do you think it is necessary?

horacimacias: if FormatWithEnv is already tested I don't see the need of testing this again (which is already partially tested by resolving MYVAR on the test)
If you have any suggestions I'm happy to change whatever needed.

tsaikd:
Can you rebase to master for checking CI state?

tsaikd:
@horacimacias In fact, it's better to use git rebase instead of creating a new PR. A new PR will lose all discussion. If you do not familiar with git command, you'd better to learn more about it. For cooperation and for career.

horacimacias:
sorry I didn't think of the history/discussion being lost. If you have any good resources to read on rebase/squash please let me know and I'll get up to speed.

PR #85, opened 2019-04-14
Allow overriding "message" field, like all other fields.

horacimacias:
Allow and use overridden "message" field, like any other field.
Otherwise things like grok filtering an overriding the "message" field do not work as the original message field keeps being used

### DIFF
--- a/config/logevent/logevent.go
+++ b/config/logevent/logevent.go
@@ -159,6 +159,12 @@ func (t LogEvent) GetValue(field string) (interface{}, bool) {
 }
 
 func (t *LogEvent) SetValue(field string, v interface{}) bool {
+	if field == "message" {
+		if value, ok := v.(string); ok {
+			t.Message = value
+			return false
+		}
+	}
 	if t.Extra == nil {
 		t.Extra = map[string]interface{}{}
 	}
--- a/config/logevent/logevent_test.go
+++ b/config/logevent/logevent_test.go
@@ -299,3 +299,21 @@ func Benchmark_Marshal_StdJSON(b *testing.B) {
 		json.Marshal(jsonMap)
 	}
 }
+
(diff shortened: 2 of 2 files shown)


tsaikd:
I think it's better to keep original message. Grok filter can parse to other fields.

tengattack:
Maybe you could update input for parsing message field to `logevent.Message`?

horacimacias:
Perhaps I'm not following @tengattack's suggestion.

The issue I had was because I was using something like the following filter config:
1) Filter for field "message" matching "%{LOGLEVEL:severity} %{DATE_FORMAT_SLASHES:date} %{JAVACLASS:java_class}:%{NUMBER:line}: %{GREEDYDATA:message}"
2) Filter for field "message" matching "something else"

For a message that matched 1, I was expecting 'message' value to change (in the same way all other fields (E.g. severity, data, java_class, line) did change). The problem is the extra field for "message" is set but is never read as Get and GetString return event.Message
I can also avoid using "message" on my grok expression and not change anything; in this case I'd suggest at least having this documented and perhaps even a log warning if a capture expression is capturing on "message" as we know it will not be modified.

tsaikd:
If you really want to change the original message, it's better to fix in SetValue().

horacimacias:
I've updated the code to now overwrite message on SetValue as suggested.

tengattack:
You should force push to your branch with one commit only :)
Here we have too much commits:
image

config/logevent/logevent.go
```diff
@@ -126,12 +126,13 @@ func (t LogEvent) Get(field string) (v interface{}) {
 	switch field {
 	case "@timestamp":
 		v = t.Timestamp
-	case "message":
```
tsaikd: You already handle `message` in SetValue(), so Get() should return from `message` property.

PR #88, opened 2019-04-16
Resolve environment property values in elasticsearch URL config

horacimacias:
Trying again pull request.

### DIFF
--- a/output/elastic/outputelastic.go
+++ b/output/elastic/outputelastic.go
@@ -13,7 +13,7 @@ import (
 	"github.com/tsaikd/gogstash/config"
 	"github.com/tsaikd/gogstash/config/goglog"
 	"github.com/tsaikd/gogstash/config/logevent"
-	elastic "gopkg.in/olivere/elastic.v6"
+	"gopkg.in/olivere/elastic.v6"
 )
 
 // ModuleName is the name used in config file
@@ -22,7 +22,8 @@ const ModuleName = "elastic"
 // OutputConfig holds the configuration json fields and internal objects
 type OutputConfig struct {
 	config.OutputConfig
-	URL             []string `json:"url"`               // elastic API entrypoints
+	URL             []string `json:"url"` // elastic API entrypoints
+	resolvedURLs    []string // URLs after resolving environment vars
 	Index           string   `json:"index"`             // index name to log
 	DocumentType    string   `json:"document_type"`     // type name to log
 	DocumentID      string   `json:"document_id"`       // id to log, used if you want to control id format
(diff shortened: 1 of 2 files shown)


output/elastic/outputelastic.go
```diff
@@ -105,8 +105,15 @@ func InitHandler(ctx context.Context, raw *config.ConfigRaw) (config.TypeOutputC
 	// map Printf to error level
 	logger := &errorLogger{logger: goglog.Logger}
 
+	// replace env var names with values on URL config
+	var resolvedURLs []string
+	for _, url := range conf.URL {
+		newURL := logevent.FormatWithEnv(url)
+		resolvedURLs = append(resolvedURLs, newURL)
```
tsaikd: In order to check resolvedURLs in unit test, you can save them in OutputConfig struct as a new field.

horacimacias:
I've updated the code and test, can you please check now?

output/elastic/outputelastic.go
```diff
@@ -21,6 +21,7 @@ const ModuleName = "elastic"
 type OutputConfig struct {
 	config.OutputConfig
 	URL             []string `json:"url"`               // elastic API entrypoints
+	ResolvedURLs    []string `json:"-"`                 // URLs after resolving environment vars
```
tsaikd: The field is generated by code, so it should be a private field (lowercase prefix), except you want it can be configured in yaml.

tengattack:
ummmm, you have conflicting files now

horacimacias:
sorry didn't see that. hopefully it's better now

output/elastic/outputelastic.go
```diff
@@ -112,12 +113,19 @@ func InitHandler(ctx context.Context, raw *config.ConfigRaw) (config.TypeOutputC
 	// map Printf to error level
 	logger := &errorLogger{logger: goglog.Logger}
 
+	// replace env var names with values on URL config
+	for _, url := range conf.URL {
+		newURL := logevent.FormatWithEnv(url)
+		conf.resolvedURLs = append(conf.resolvedURLs, newURL)
```
tengattack: Do we need to reset it before appending urls?

horacimacias: conf.resolvedURLs ? 

I don't think so as this comes empty after creating conf in 

```suggestion
conf := DefaultOutputConfig()
err := config.ReflectConfig(raw, &conf)
```
and ReflectConfig does not populate resolvedURLs as it's private.

tengattack: OK

tsaikd:
Looks like some codes not pass go fmt.

horacimacias:
ok sorry. I've pushed an update after `go fmt` but now I think there are more commits than needed. Is this ok now or do I need to rebase somehow?

output/elastic/outputelastic.go
```diff
@@ -22,7 +22,8 @@ const ModuleName = "elastic"
 // OutputConfig holds the configuration json fields and internal objects
 type OutputConfig struct {
 	config.OutputConfig
-	URL             []string `json:"url"`               // elastic API entrypoints
+	URL             []string `json:"url"` // elastic API entrypoints
```
tengattack: We should align comment here

horacimacias: the current format is what `go fmt` produces. If I run `go fmt` again I see no changes and if I modify manually and run `go fmt`again the changes are undone.

Are you suggesting to align all comments in the OutputConfig struct?

I can do that for sure, just that I don't think the code will pass `go fmt` as before

Issue #80, opened 2019-03-25
conditional branching?


horacimacias:
I'm trying to migrate from logstash to gogstash as the memory usage of my environment is getting out of hand.
I couldn't find anything regarding conditional branching on input/filter/output elements like logstash does. Is there any similar thing in gogstash?

tengattack:
This is what I'ms using in production:
```yml
chsize: 5000
worker: 4
event:
  sort_map_keys: true
  remove_field: ['@metadata']

input:
  - type: beats
    port: 5065
    reuseport: true
  - type: socket
    socket: udp
    address: '0.0.0.0:5066'
    reuseport: true
    buffer_size: 65536

filter:
  - type: cond
    condition: "[fileset.module] == 'nginx' && [fileset.name] == 'access'"
    filter:
      - type: grok
        match:
          - "%{IPORHOST:nginx.access.remote_ip} - %{DATA:nginx.access.user_name} \\[%{HTTPDATE:nginx.access.time}(?:| %{NUMBER:nginx.access.timestamp})\\
\] \"%{DATA:nginx.access.host}\" \"%{WORD:nginx.access.method} %{DATA:nginx.access.url} HTTP/%{NUMBER:nginx.access.http_version}\" %{NUMBER:nginx.accesss
.response_code:int} %{NUMBER:nginx.access.body_sent.bytes} \"%{DATA:nginx.access.referrer}\" \"%{DATA:nginx.access.agent}\" \"(?:-|%{DATA:nginx.access.xx
_forwarded_for})\" ?%{NUMBER:nginxx
.access.request_time:float} (?:%{NUMBER:nginx.access.upstream_response_time:float}|-)"
        source: message
        #patterns_path: /etc/gogstash/grok-patterns
      - type: add_field
        key: 'read_timestamp'
        value: '%{@timestamp}'
      - type: cond
        condition: "!('gogstash_filter_grok_error' IN map([tags]))"
        filter:
          - type: remove_field
            remove_message: true
          - type: cond
            condition: "!empty([nginx.access.timestamp])"
            filter:
              - type: date
                format: ["UNIX"]
                source: "nginx.access.timestamp"
            else_filter:
              - type: date
                format: ["dd/MMM/YYYY:H:m:s Z"]
                joda: true
                source: "nginx.access.time"
          - type: cond
            condition: "!('gogstash_filter_date_error' IN map([tags]))"
            filter:
              - type: remove_field
[...]

PR #90, opened 2019-04-25
Add common filter logic add/remove tag/field

horacimacias:
I'm suggesting this change so all filters allow adding/removing tags and filters, without needing a specific filter for this operation

### DIFF
--- a/config/filter.go
+++ b/config/filter.go
@@ -17,11 +17,35 @@ var (
 type TypeFilterConfig interface {
 	TypeCommonConfig
 	Event(context.Context, logevent.LogEvent) logevent.LogEvent
+	CommonFilter(context.Context, logevent.LogEvent) logevent.LogEvent
+}
+
+func (f *FilterConfig) CommonFilter(ctx context.Context, event logevent.LogEvent) logevent.LogEvent {
+
+	event.AddTag(f.AddTags...)
+	event.RemoveTag(f.RemoveTags...)
+	for _, field := range f.RemoveFields {
+		event.Remove(field)
+	}
+	for _, f := range f.AddFields {
+		event.SetValue(f.Key, event.Format(f.Value))
+	}
+	return event
 }
 
 // FilterConfig is basic filter config struct
 type FilterConfig struct {
 	CommonConfig
+	AddTags      []string      `yaml:"add_tag" json:"add_tag"`
+	RemoveTags   []string      `yaml:"remove_tag" json:"remove_tag"`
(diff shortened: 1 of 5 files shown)


tsaikd:
I prefer keep every component as simple as possible, but this is an open issue for discussion. Any other opinion?

horacimacias:
The logic inside components will be the same as this common behaviour is implemented in a common/inherited method.
The benefit of this common logic is that users of gogstash will not need "extra" components to only add/remove tags/fields.
The complexity is the same; just that right now the logic is on a specific component and with this change it will be in a common place.
As a side note, this is also how logstash does it; the amount of common logic is different but add/remove tags/fields is a common feature. For example see https://www.elastic.co/guide/en/logstash/current/plugins-filters-mutate.html#plugins-filters-mutate-common-options

config/filter.go
```diff
@@ -73,6 +97,7 @@ func (t *Config) startFilters() (err error) {
 				}
 			case event := <-t.chInFilter:
 				for _, filter := range filters {
+					event = filter.CommonFilter(t.ctx, event)
```
tengattack: I think this filter should be processed only after event filter success?

tsaikd: Why? The behavior seems does not conflict.

tengattack: I think we only need action after filter success. Otherwise, we can use mutate filter do the same thing.

BTW, logstash's filters also do this on filter success:
https://www.elastic.co/guide/en/logstash/current/plugins-filters-date.html#plugins-filters-date-common-options

tengattack: So, I think we could change the `Event` function and return the `success bool` to inform `CommonFilter` to do so.

tsaikd: I'm neutral about this, @horacimacias what do you think?

tengattack: For example, I need to parse string to date, and it may not success always, I should only remove the original value when it is successful.

Like what I did in #80 's example:
```
          - type: cond
            condition: "!empty([nginx.access.timestamp])"
            filter:
              - type: date
                format: ["UNIX"]
                source: "nginx.access.timestamp"
            else_filter:
              - type: date
                format: ["dd/MMM/YYYY:H:m:s Z"]
                joda: true
                source: "nginx.access.time"
          - type: cond
            condition: "!('gogstash_filter_date_error' IN map([tags]))"
            filter:
              - type: remove_field
                fields: ['nginx.access.timestamp', 'nginx.access.time']
```

If we do this common filter only if filter success, it could be simplified without loss original value.

horacimacias: I think that makes sense, yes. Agreed

PR #97, opened 2019-05-15
Added logic to check common configuration

horacimacias:
This change allows mutate filter to be configured only with common config parameters.
E.g.

  - type: mutate
    add_tag: ["testing"]

there is also an IsConfigured function that can be used by filters to validate

### DIFF
--- a/config/filter.go
+++ b/config/filter.go
@@ -20,6 +20,11 @@ type TypeFilterConfig interface {
 	CommonFilter(context.Context, logevent.LogEvent) logevent.LogEvent
 }
 
+// IsConfigured returns whether common configuration has been setup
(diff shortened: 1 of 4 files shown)


tengattack:
It seems conflicted.

config/filter.go
```diff
@@ -20,6 +20,11 @@ type TypeFilterConfig interface {
 	CommonFilter(context.Context, logevent.LogEvent) logevent.LogEvent
 }
 
+// IsConfigured returns whether common configuration has been setup
+func (f *FilterConfig) IsConfigured() bool {
```
tsaikd: Can you add unit test for this function?

horacimacias: Sure. done

PR #99, opened 2019-05-15
Fixed issues when writing to existing dir new file

horacimacias:
This fixes a couple issues:
* Using a new file on existing directory: file needs to be opened with Create permissions (not append)
* Receiving events on a file that existed but was deleted. File is created again (if configured) and event is written to newly created file. Problem was the event was not being re-written into newly created file

### DIFF
--- a/output/file/outputfile_test.go
+++ b/output/file/outputfile_test.go
@@ -8,9 +8,8 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
-
 	"github.com/golang/mock/gomock"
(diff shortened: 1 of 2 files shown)


output/file/outputfile_test.go
```diff
@@ -2,13 +2,14 @@ package outputfile
...
 	"context"
-	"github.com/stretchr/testify/assert"
 	"os"
 	"strconv"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+
```
tsaikd: Unnecessary empty line

PR #100, opened 2019-05-15
Changed log level when no match

horacimacias:
Suggesting this change as often grok is used to set tags and make branching decisions based on that. Logging as error is just polluting the output too much


tsaikd:
@tengattack What do you think about this PR?

tengattack:
I aggree with that

PR #104, opened 2019-06-06
Fields created using json filter should be parseable with the date filter

horacimacias:
jsonfilter unmarshalls a string into json. 
By default, json numbers will be unmarshalled to float64.
Date filter is not expecting numbers in exponential notation such as float64's Printf("%v")

### DIFF
--- a/filter/date/filterdate_test.go
+++ b/filter/date/filterdate_test.go
@@ -2,6 +2,8 @@ package filterdate
 
 import (
 	"context"
+	"fmt"
+	filterjson "github.com/tsaikd/gogstash/filter/json"
 	"strings"
 	"testing"
 	"time"
@@ -17,6 +19,7 @@ import (
 func init() {
 	goglog.Logger.SetLevel(logrus.DebugLevel)
 	config.RegistFilterHandler(ModuleName, InitHandler)
+	config.RegistFilterHandler(filterjson.ModuleName, filterjson.InitHandler)
 }
 
 func Test_filter_date_module(t *testing.T) {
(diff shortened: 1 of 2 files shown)


tsaikd:
CI failed, please format your go source codes.

filter/date/filterdate_test.go
```diff
@@ -174,3 +177,111 @@ filter:
...
+  - type: json
+    source: json
+  - type: date
+    format: ["UNIX"]
+    source: time_local
+	`)))
+	require.NoError(err)
+	require.NoError(conf.Start(ctx))
+
+	require.NoError(err)
```
tsaikd: ???

filter/date/filterdate_test.go
```diff
@@ -174,3 +177,111 @@ filter:
...
+		{
+			name:  "int",
+			input: int64(1558348989869),
+			sec:   int64(1558348989869),
+			nsec:  int64(0),
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			r := require.New(t)
```
tsaikd: Line 217 defined?

tsaikd: I think it's better to move `r` definition to line 215. So it can avoid initializing in a for loop.

tengattack: I think here has a new `*tesing.T` pass in line 267, it has to be defined here

tsaikd: oh...you are right.

PR #110, opened 2019-06-28
support TLS for httplisten

tao-sj:
add README
support https


tsaikd:
How about add an unit test to make sure your TLS config will work?

tao-sj:
> How about add an unit test to make sure your TLS config will work?

done. test passed.

PR #114, opened 2019-08-01
Add event ok return to indicate common filter

tengattack:
By discussing at #90, common filter will be processed only after event filter success.


tsaikd:
@tengattack test cases do not pass

PR #117, opened 2019-08-13
http: support server authenticating client

tao-sj:
enable https client authentication if users configures CA.


input/httplisten/httplisten_test.go
```diff
@@ -101,3 +101,47 @@ input:
...
+	ctx := context.Background()
+	conf, err := config.LoadFromYAML([]byte(strings.TrimSpace(`
+debugch: true
+input:
+  - type: httplisten
+    address: "127.0.0.1:8989"
+    path: "/tls2/"
+    cert: "./leaf.pem"
+    key:  "./leaf.key"
+    CA:   "./root.pem"
```
tengattack: ca should be lowercase

input/httplisten/httplisten.go
```diff
@@ -85,7 +89,43 @@ func (i *InputConfig) Start(ctx context.Context, msgChan chan<- logevent.LogEven
...
+
+			if i.CA != "" {
+				// enable client certificate
+				f, err := os.Open(i.CA)
+				if err != nil {
+					logger.Fatal(err)
+					return
+				}
+				content, err := ioutil.ReadAll(f)
+				f.Close()
```
tsaikd: at least log error?

input/httplisten/httplisten.go
```diff
@@ -85,7 +89,43 @@ func (i *InputConfig) Start(ctx context.Context, msgChan chan<- logevent.LogEven
...
+					ClientAuth:   tls.RequireAndVerifyClientCert,
+					Certificates: []tls.Certificate{srvCert},
+				}
+			} else {
+				tlsConfig = &tls.Config{
+					Certificates: []tls.Certificate{srvCert},
+				}
+			}
+			l, err := tls.Listen("tcp", i.Address, tlsConfig)
+			if err == nil {
```
tsaikd: variable err shadow here, so it will not return err even err != nil

input/httplisten/httplisten_test.go
```diff
@@ -101,3 +101,47 @@ input:
...
+	assert.NotNil(assert)
+	require := require.New(t)
+	require.NotNil(require)
+
+	ctx := context.Background()
+	conf, err := config.LoadFromYAML([]byte(strings.TrimSpace(`
+debugch: true
+input:
+  - type: httplisten
+    address: "127.0.0.1:8989"
```
tsaikd: You listen the same port with `Test_input_httpslisten_module`, and listen error doesn't return. So you pass the unit test...Q_Q

input/httplisten/httplisten.go
tao-sj: error is logged below - line 109

input/httplisten/httplisten_test.go
tao-sj: good catch! though it'll not happen in real world.
I was thinking this feature should have service close API so that we don't need to listen on different ports in UT.

input/httplisten/httplisten.go
tsaikd: I mean `f.Close()` will return error, but you ignored.

tsaikd:
Why generating a new key pair `server.x` instead of the old `leaf.x`?

tao-sj:
Old key will expire in one year, new key is 10 years
, since we have client key pair now, renaming to server is more clear.

input/httplisten/httplisten_test.go
```diff
@@ -101,3 +101,59 @@ input:
...
+	assert.NotNil(roots)
+	ok := roots.AppendCertsFromPEM(rootPEM)
+	assert.Equal(ok, true)
+
+	clientCert, err := tls.LoadX509KeyPair("./client.pem", "./client.key")
+	require.NoError(err)
+
+	tlsConfig := tls.Config{
+		RootCAs:      roots,
+		Certificates: []tls.Certificate{clientCert},
```
tsaikd: Please add one more test case about client without certificate.

input/httplisten/httplisten_test.go
```diff
@@ -101,3 +101,65 @@ input:
...
+	assert.NotNil(roots)
+	ok := roots.AppendCertsFromPEM(rootPEM)
+	assert.Equal(ok, true)
+
+	clientCert, err := tls.LoadX509KeyPair("./client.pem", "./client.key")
+	require.NoError(err)
+
+	// case 1: without client cert
+	tlsConfig := tls.Config{
+		RootCAs:      roots,
```
tsaikd: Error: found unformatted go source: "/home/travis/gopath/src/github.com/tsaikd/gogstash/input/httplisten/httplisten_test.go"

tao-sj: hmm, yeah, it needs reformatted

PR #122, opened 2020-02-13
Added Codec support to a number of modules

aantono:


tsaikd:
Unit tests failed...@_@

aantono:
So trying to debug the test failure, and so far got down to the double slash escaping failing the json parser, but can't understand why:
`match: "%{SPACE}%{GREEDYDATA:timestamp} \\[%{LOGLEVEL:loglevel}] %{DATA}"` the use of `\\` to escpae `[` causes json unmarshal to fail.  But if I don't escape the `[`, that causes regex matcher to not treat `[` as a character, and instead treat it as a character block. 🤷‍♂

PR #123, opened 2020-03-26
add kafka input support

McEchean:
add kafka input support

### DIFF
--- a/input/kafka/inputkafka.go
+++ b/input/kafka/inputkafka.go
@@ -1,0 +1,202 @@
+package inputkafka
+
+import (
+	"context"
+	"github.com/Shopify/sarama"
+	"github.com/tsaikd/gogstash/config"
+	"github.com/tsaikd/gogstash/config/goglog"
+	"github.com/tsaikd/gogstash/config/logevent"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+)
+
+// ModuleName is the name used in config file
+const ModuleName = "kafka"
+
+// InputConfig holds the configuration json fields and internal objects
+type InputConfig struct {
+	config.InputConfig
+	Version      string   `json:"version"`       // Kafka cluster version, eg: 0.10.2.0
+	Brokers      []string `json:"brokers"`       // Kafka bootstrap brokers to connect to, as a comma separated list
+	Topics       []string `json:"topics"`        // Kafka topics to be consumed, as a comma seperated list
+	Group        string   `json:"group"`         // Kafka consumer group definition
+	OffsetOldest bool     `json:"offset_oldest"` // Kafka consumer consume initial offset from oldest
+	Assignor     string   `json:"assignor"`      // Consumer group partition assignment strategy (range, roundrobin)
+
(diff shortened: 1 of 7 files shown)


input/kafka/inputkafka.go
```diff
@@ -0,0 +1,200 @@
...
+	"syscall"
+)
+
+// ModuleName is the name used in config file
+const ModuleName = "kafka"
+
+// InputConfig holds the configuration json fields and internal objects
+type InputConfig struct {
+	config.InputConfig
+	Version      string   `json:"version"`       //Kafka cluster version, eg: 2.1.1
```
tengattack: We should add a space after `//`, like `// InputConfig holds ...`

input/kafka/inputkafka.go
```diff
@@ -0,0 +1,200 @@
...
+			// recreated to get the new claims
+			if err := client.Consume(ctx, t.Topics, &cum); err != nil {
+				goglog.Logger.Errorf("Error from consumer: %v", err)
+			}
+			// check if context was cancelled, signaling that the consumer should stop
+			if ctx.Err() != nil {
+				return
+			}
+
+			cum.ready = make(chan bool)
```
tsaikd: Why renew the ready channel? (just curious)

input/kafka/inputkafka.go
```diff
@@ -0,0 +1,200 @@
...
+	// Do not move the code below to a goroutine.
+	// The `ConsumeClaim` itself is called within a goroutine, see:
+	// https://github.com/Shopify/sarama/blob/master/consumer_group.go#L27-L29
+	for message := range claim.Messages() {
+		//goglog.Logger.Printf("Message claimed: value = %s, timestamp = %v, topic = %s", string(message.Value), message.Timestamp, message.Topic)
+		var extra = map[string]interface{}{
+			"topic": message.Topic,
+			"timestamp": message.Timestamp,
+		}
+		ok, err := c.i.Codec.Decode(context.TODO(), string(message.Value), extra, []string{}, c.ch)
```
tsaikd: Why use `context.TODO()`?
1. why not main context?
2. why not `context.Background()`?

input/kafka/inputkafka.go
McEchean: the example of sarama written like this,I think the reason is this. In the `Setup` func, it was originally written `close (c.ready)`, so need to renew the ready channel, but I changed it to `c.ready <-true` in the setup, and I will try to It changed back.

input/kafka/inputkafka.go
McEchean: changed to main context

input/kafka/inputkafka.go
tsaikd: I guess the example assumes you would reuse the channel. But in this case, the `Start()` executes only once. Anyway, renew the channel might not break anything.

tsaikd:
Could you squash the commits before I merge it?

McEchean:
> Could you squash the commits before I merge it?

Ok, it's done

Issue #128, opened 2020-04-18
Elastic output document type as deprecated


pachico:
I still need this feature for a 6.4.2 ES cluster but it seems to be deprecated.
Would it make sense to have an "elastic_legacy" output plugin that still supports it?
Anything to take in consideration?

pachico:
Closed as v5 has been included again

PR #130, opened 2020-04-19
Elastic legacy

pachico:
This change should allow this plugin to work in new and old Elasticsearch versions. 
What do you think about it?

### DIFF
--- a/output/elastic/README.md
+++ b/output/elastic/README.md
@@ -14,9 +14,9 @@ output:
     # index name to log
     index: "gogstash-index-test"
 
-    # (deprecated)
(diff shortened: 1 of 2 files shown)


tsaikd:
https://github.com/tsaikd/gogstash/tree/0.1.21/output/elasticv5

Add the v5 module back. You can close #130 and #128, If it's solved your problem.

pachico:
Thanks a lot, sir

aantono:
Does this work with Elastic 6.x as well as the 5.x lines?

tsaikd:
I'm not sure, you need to test by yourself for the dedicated features.

PR #132, opened 2020-04-20
[WIP]  Output clickhouse

pachico:
Hi, this is still work in progress (features, docs and tests still missing) but basic functionality is there.
Do you think you can have a look at it and let me know if I'm totally misguided or if you see this as a possible merge in the future?

Thanks in advance


output/clickhouse/clickhouse.go
```diff
@@ -0,0 +1,120 @@
...
+	newEventAsJSON, err := jsoniter.Marshal(newEventMap)
+
+	if err != nil {
+		return err
+	}
+
+	queryString := fmt.Sprintf("INSERT INTO %s FORMAT JSONEachRow", t.TargetTable)
+	requestURL := fmt.Sprintf("%s?query=%s", t.DSN, url.QueryEscape(queryString))
+
+	request, _ := http.NewRequest("POST", requestURL, strings.NewReader(string(newEventAsJSON)))
```
tsaikd, line 101: Please do not ignore any errors.

output/clickhouse/clickhouse.go
```diff
@@ -0,0 +1,120 @@
...
+// @todo: create bulk functionality
+// @todo: add gzip functionality
+func (t *OutputConfig) Output(ctx context.Context, event logevent.LogEvent) (err error) {
+
+	newEventMap := make(map[string]interface{})
+
+	for _, key := range t.Fields {
+		fieldValue, exists := event.GetValue(key)
+
+		if true == exists {
```
tsaikd, line 77:
```suggestion
		if exists {
```

output/clickhouse/clickhouse.go
```diff
@@ -0,0 +1,120 @@
+package outputclickhouse
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"log"
```
tsaikd, line 7: try `github.com/tsaikd/gogstash/config/goglog`

output/clickhouse/clickhouse.go
```diff
@@ -0,0 +1,120 @@
...
+
+	request, _ := http.NewRequest("POST", requestURL, strings.NewReader(string(newEventAsJSON)))
+	resp, err := t.httpClient.Do(request)
+
+	if err != nil {
+		log.Println("Error executing request to Clickhouse: ", err)
+	}
+
+	if resp.StatusCode != 200 {
+		defer resp.Body.Close()
```
tsaikd, line 109: https://godoc.org/net/http
> The client must close the response body when finished with it

Does the `pester` HTTP client has different behavior?

output/clickhouse/clickhouse.go
pachico: Sure thing, will do.

output/clickhouse/clickhouse.go
pachico: I see not all plugins use it but if it's the desired one I will implement it.

output/clickhouse/clickhouse.go
pachico: Yes, and I want to implement it. It mostly handles retries, which reduces the possibility of missed documents. I want users to be able to customise it as they do for elastic plugin.

pachico:
I addition, I want to implement batch processing since ClickHouse prefers it

dairongpeng:
Hi, is there a plan to merge?

tsaikd:
I don't see any actions after the last review.

pachico:
I won't proceed with this draft since I'm currently using another solution that already has ClickHouse output option.

PR #137, opened 2020-08-19
Add sensible default useragent regexes; add README

### LINKED ISSUE #136: useragent filter should use uap-go regexes by default
I'm trying to use the `useragent` filter and getting the error: `filter useragent 'regexes' not configured`. I would expect gogstash to default to using the `DefinitionYaml` variable [available in uap-go](https://github.com/ua-parser/uap-go/blob/e1c09f13e2fe20d591e97106ab10758f5a80f07b/uaparser/yaml.go#L3), so I don't have to configure this unless I require a different set of regexes.

I'd be happy to create PR for this if it sounds reasonable.

edsrzf:
This adds a sensible default configuration to the `useragent` filter when no regex file is specified, using [`uaparser.NewFromSaved`](https://godoc.org/github.com/ua-parser/uap-go/uaparser#NewFromSaved). It also adds a README for this filter.

Closes #136.

### DIFF
--- a/filter/useragent/README.md
+++ b/filter/useragent/README.md
@@ -1,0 +1,59 @@
+gogstash useragent filter module
+=============================
+
+This filter parses a useragent string in the log event message to a structured form.
+
+## Synopsis
+
+```yaml
+filter:
+  - type: useragent
+    # message field to parse
+    source: headers.user_agent
+    # target field
+    target: user_agent
+    # (optional) YAML file of user agent regexes to match. Default: regexes from uaparser
+    # You can find the latest version of this here:
+    # <https://github.com/ua-parser/uap-core/blob/master/regexes.yaml>
+    regexes: "./regex.yaml"
+    # (optional) Number of entries to cache, to avoid re-parsing the same user agents repeatedly. Default: 100000
+    cache_size: 1000
+```
+
+## Example for useragent
+
+Input string
+```json
+{
+    "headers": {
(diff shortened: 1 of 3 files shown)


filter/useragent/filteruseragent.go
```diff
@@ -15,6 +15,7 @@ const ModuleName = "useragent"
 
 // errors
 var (
+	// Deprecated: this error is never returned.
```
edsrzf, line 18: Not sure how concerned we are about backward compatibility and keeping this around. Happy to remove this if it's not an issue.

tsaikd: I think it's OK to remove unused variables.

filter/useragent/README.md
```diff
@@ -0,0 +1,59 @@
...
+    cache_size: 1000
+```
+
+## Example for useragent
+
+Input string
+```json
+{
+    "headers": {
+      "user_agent": "Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/59.0.3071.115 Safari/537.36"
```
tengattack, line 29: Maybe we can keep consistent indent

filter/useragent/README.md
```diff
@@ -0,0 +1,59 @@
...
+  - type: useragent
+    source: headers.user_agent
+    target: user_agent
+```
+
+Produces the following output:
+```json
+{
+    "headers": {
+      "user_agent": "Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/59.0.3071.115 Safari/537.36"
```
tengattack, line 46: Same as above

tengattack (approved):

Issue #144, opened 2021-06-12
How to log in to elasticsearch


helgeolav:
Hi

I am looking into the output module elastic and I can't see a way to send in some kind of authentication. How can I configure a username/password to log into Elasticsearch?

tsaikd:
Thanks for your PR #143

helgeolav:
If there is no obvious answer to this I can create a new PR with a solution to this. I seem to need a few more changes to the elastic output as I also need to use the simple client. Behind NAT/Docker etc there are some issues when the client tries to be smart and choose endpoint for itself even without sniff.

PR #145, opened 2021-06-15
Updates to Elastic

helgeolav:
This is an update to #143 where I also added the username/auth to elasticv5 and also added support for NewSimpleClient as mentioned in #144. NewSimpleClient solves a few connectivity issues with some installations. Read more about it [here](https://github.com/olivere/elastic/wiki/Connection-Problems#how-to-figure-out-connection-problems).

### DIFF
--- a/output/elastic/README.md
+++ b/output/elastic/README.md
@@ -34,6 +34,10 @@ output:
     # find all nodes of your cluster, https://github.com/olivere/elastic/wiki/Sniffing
     sniff: true
 
+    # (optional) default: false
+    # uses the simpleclient connection type instead of the normal client, helpful in cases you get error messages even when the client can connect
+    simple_client: true
+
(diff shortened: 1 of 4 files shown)


tsaikd:
Please reorganize your commits and rebase to master.

helgeolav:
Hi

Now I am a bit lost. I have tried to rebase but I don't make anything happen. Can you tell in more detail what I need to do and how?

It seems it could be easier to just delete my fork, do a new fork and then apply the changes. But I have read that it is a bad practice.

tsaikd:
git rebase is a little bit advanced skill, and it's helpful when you are cleaning the dirty history when developing. If you are a mac user, there is a excellent tool [GitUp](https://github.com/git-up/GitUp) makes git rebase easier. But you should know exactly what it does for rebasing to avoid any disasters.

Issue #146, opened 2021-06-15
Input - UDP packet mode


helgeolav:
I am missing a way to handle packet based messages in UDP. (Think syslog etc.) Input socket only handles line delimited inputs, and this does not work well with our input. I have been looking through the code and I think the best way to solve this is to create a new UDP-only input type. (The alternative I looked into is to change socket to be more packet-like for UDP connections.)

In addition I could also add support for syslog parsing as a codec, filter or both using [go-syslog](https://github.com/influxdata/go-syslog).

Will this be an acceptable solution?

tengattack:
try:
https://github.com/tsaikd/gogstash/tree/master/input/socket

helgeolav:
The problem with socket is in the first line of the documentation: Input event message should end with new line (\n).

If UDP packets arrive and does not have \n in them gogstash will just keep them and fill up the memory. But it does not produce any output. Syslog messages does not end with \n.

The reason for me suggesting to write a new input is that socket is stream driven and I will need to make several changes to make it packet-oriented. My plan was to use net.PacketConn interface so I also can collect the source IP for each packet and add it as a field.

helgeolav:
Implemented in #152.

PR #148, opened 2021-06-26
Support for UDP packet mode

helgeolav:
As discussed in #146 i have made some changes to the code to handle UDP packets in a packet-oriented way instead of the stream-oriented way that the socket input handles UDP traffic. By setting a new configuration option "packetmode" to true each packet is processed separately and sent in for further processing. A field "host_ip" is also added that is the source IP/port for the packet.

### DIFF
--- a/input/socket/inputsocket.go
+++ b/input/socket/inputsocket.go
@@ -31,6 +31,8 @@ type InputConfig struct {
 	Address    string `json:"address"`
 	ReusePort  bool   `json:"reuseport"`
 	BufferSize int    `json:"buffer_size"`
+	// packetmode is only valid for UDP sessions and handles each packet as a message on its own
+	PacketMode bool `json:"packetmode"`
 }
 
 // DefaultInputConfig returns an InputConfig struct with default values
@@ -121,7 +123,11 @@ func (i *InputConfig) Start(ctx context.Context, msgChan chan<- logevent.LogEven
 		if err != nil {
 			return err
 		}
-		return i.handleUDP(ctx, conn, msgChan)
+		if i.PacketMode {
+			return i.handleUDPpacketMode(ctx, conn, msgChan)
+		} else {
+			return i.handleUDP(ctx, conn, msgChan)
+		}
 	default:
 		return ErrorUnknownSocketType1.New(nil, i.Socket)
 	}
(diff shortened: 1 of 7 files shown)


input/socket/inputsocket.go
```diff
@@ -121,7 +123,11 @@ func (i *InputConfig) Start(ctx context.Context, msgChan chan<- logevent.LogEven
 		if err != nil {
 			return err
 		}
-		return i.handleUDP(ctx, conn, msgChan)
+		if i.PacketMode == false {
+			return i.handleUDP(ctx, conn, msgChan)
+		} else {
+			return i.handleUDPpacketMode(ctx, conn, msgChan)
+		}
```
tsaikd, line 130:
```suggestion
		if i.PacketMode {
			return i.handleUDPpacketMode(ctx, conn, msgChan)
		} else {
			return i.handleUDP(ctx, conn, msgChan)
		}
```

input/socket/inputsocket.go
```diff
@@ -31,6 +31,8 @@ type InputConfig struct {
 	Address    string `json:"address"`
 	ReusePort  bool   `json:"reuseport"`
 	BufferSize int    `json:"buffer_size"`
+	// packetmode is only valid for UDP sessions and handles each packet as a message on its own
+	PacketMode bool `json:"packetmode"`
```
tsaikd, line 35:
```suggestion
	PacketMode bool `json:"packet_mode"`
```

input/socket/inputsocket.go
```diff
@@ -152,6 +158,54 @@ func (i *InputConfig) Start(ctx context.Context, msgChan chan<- logevent.LogEven
...
+				}
+			}
+			// handle error
+			if err == io.EOF {
+				break
+			} else if err != nil {
+				return err
+			}
+		}
+		return nil
```
tsaikd, line 203:
```suggestion
			switch err {
			case nil:
				// continue processing
			case io.EOF:
				return nil
			default:
				return err
			}
```

input/socket/inputsocket.go
```diff
@@ -152,6 +158,54 @@ func (i *InputConfig) Start(ctx context.Context, msgChan chan<- logevent.LogEven
...
+			if n > 0 {
+				if n >= i.BufferSize {
+					const bufIncSize = 600
+					logger.Errorf("%v UDP receive buffers (%v bytes) too small, should be increased!", i.Address, i.BufferSize)
+					i.BufferSize += bufIncSize
+					b = make([]byte, i.BufferSize)
+				} else {
+					extras := make(map[string]interface{})
+					extras["host_ip"] = addr.String()
+					_, _ = i.Codec.Decode(ctx, b[:n], extras, []string{}, msgChan)
```
tsaikd, line 193: ignore all decode errors?

input/socket/inputsocket.go
```diff
@@ -152,6 +158,54 @@ func (i *InputConfig) Start(ctx context.Context, msgChan chan<- logevent.LogEven
...
+			// handle data
+			if n > 0 {
+				if n >= i.BufferSize {
+					const bufIncSize = 600
+					logger.Errorf("%v UDP receive buffers (%v bytes) too small, should be increased!", i.Address, i.BufferSize)
+					i.BufferSize += bufIncSize
+					b = make([]byte, i.BufferSize)
+				} else {
+					extras := make(map[string]interface{})
+					extras["host_ip"] = addr.String()
```
tsaikd, line 192:
```suggestion
					extras := map[string]interface{}{
						"host_ip": addr.String(),
					}
```

input/socket/inputsocket.go
helgeolav: I based my code on the existing code. All other code paths also ignore the result from Decode. I only added _, _ to be explicit about me ignoring the result.

If I should handle this I am a bit unsure how to handle the result. Please advice. I also should handle the error in the other code paths as well?

As Decode returns (ok bool, err error) I have four possible results:

```text
ok err
======
fa nil message not sent, no error to explain why
fa err message not sent with error
tr nil everything is ok, normal behaviour
tr err message sent, with error
```
I could write some code to send a log message in case the message was not sent or there is an error?

aantono: The method signature is more nuanced.
The real failure is ONLY when `err` is returned.  That's a clear indication that the processing of data input has failed, either with parsing or with some other, codec-specific, data processing/decoding.
If `ok` is false, that only means that a new `Event` struct was not sent to the `msgChan`.  This could happen IFF a particular Codec implementation does not do any `Event` decoding per-say, which results in a new `Event`, but rather extracts some of the data and either does some "pre-processing" or adds attributes to the `Context` object.  (Neither of those are common/typical cases, but theoretically they both could occur)

In your particular case, I would only check for `err != nil` as a check.

helgeolav:
I have updated my fork with all the changes that have been discussed up to now. I also added error handling for Codec in parse() so the behaviour is similar however this input is used.

input/socket/inputsocket.go
```diff
@@ -152,6 +158,61 @@ func (i *InputConfig) Start(ctx context.Context, msgChan chan<- logevent.LogEven
...
+			switch err {
+			case nil:
+			// continue processing
+			case io.EOF:
+				return nil
+			default:
+				return err
+			}
+		}
+		return nil
```
tsaikd, line 210:
```suggestion
```

unreachable code

helgeolav: I have removed line 210 now. Please accept change.

tsaikd:
Sorry for replying lately, please rearrange your commits before merging.

helgeolav:
Closing, change was implemented at #152.

PR #149, opened 2021-06-27
Refactored DecodeEvent and Decode to remove duplicate logic 

aantono:
Made the method purposes cleaner (removing tight coupling to JSONDecoder signature from Beats)
Extracted common functionality for `event.Extra` processing into a shared method - `populateEventExtras`

### DIFF
--- a/input/beats/inputbeats.go
+++ b/input/beats/inputbeats.go
@@ -104,7 +104,23 @@ func (t *InputConfig) Start(ctx context.Context, msgChan chan<- logevent.LogEven
 			l = tls.NewListener(l, t.tlsConfig)
 		}
 		return l, err
-	}, addr, server.JSONDecoder(t.Codec.DecodeEvent))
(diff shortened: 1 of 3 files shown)


input/beats/inputbeats.go
```diff
@@ -104,7 +104,33 @@ func (t *InputConfig) Start(ctx context.Context, msgChan chan<- logevent.LogEven
...
+
+		//  switch e := v.(type) {
+		//	case *interface{}:
+		//		*e = event
+		//	case *logevent.LogEvent:
+		//		*e = event
+		//	default:
+		//		return config.ErrorUnsupportedTargetEvent
+		//	}
+		//	return nil
```
tsaikd, line 130: any reason for keeping comments?

aantono: my bad, forgot to remove them before commit. ;)

PR #162, opened 2021-07-29
filter hash

helgeolav:
This is a proposal for a hash filter. The idea about making a hash is that if a source (or input) has "at least once delivery" and the destination (or output) can drop duplicates then you use a hash to accomplish this.

### DIFF
--- a/filter/hash/filterhash.go
+++ b/filter/hash/filterhash.go
@@ -1,0 +1,191 @@
+package hash
+
+import (
+	"context"
+	"crypto/md5"
+	"crypto/sha1"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"github.com/tsaikd/gogstash/config"
+	"github.com/tsaikd/gogstash/config/goglog"
+	"github.com/tsaikd/gogstash/config/logevent"
+	"hash"
+	"hash/adler32"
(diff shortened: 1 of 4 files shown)


filter/hash/filterhash.go
```diff
@@ -0,0 +1,196 @@
...
+			if v.Name == conf.Kind {
+				conf.hash32 = v.Init
+				break
+			}
+		}
+	}
+	if conf.hash == nil && conf.hash32 == nil {
+		msg := fmt.Sprint("filterhash unsupported hash algo: ", conf.Kind)
+		err = errors.New(msg)
+		return err
```
tsaikd, line 127: `return fmt.Errorf( ... )`?

filter/hash/filterhash.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// i32tob convert uint32 to byte array
+func i32tob(val uint32) []byte {
+	r := make([]byte, 4)
+	for i := uint32(0); i < 4; i++ {
+		r[i] = byte((val >> (8 * i)) & 0xff)
+	}
+	return r
+}
```
tsaikd, line 169: https://stackoverflow.com/questions/16888357/convert-an-integer-to-a-byte-array
Better?

helgeolav:
I have implemented both changes.

PR #163, opened 2021-07-29
add_field doc and overwrite

helgeolav:
I added a README to this filter and also added a new configuration field overwrite that allows it to overwrite an existing field.

### DIFF
--- a/filter/addfield/filteraddfield.go
+++ b/filter/addfield/filteraddfield.go
@@ -13,8 +13,9 @@ const ModuleName = "add_field"
 // FilterConfig holds the configuration json fields and internal objects
 type FilterConfig struct {
 	config.FilterConfig
-	Key   string `json:"key"`
-	Value string `json:"value"`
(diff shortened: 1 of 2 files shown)


filter/addfield/filteraddfield.go
```diff
@@ -40,8 +41,10 @@ func InitHandler(ctx context.Context, raw *config.ConfigRaw) (config.TypeFilterC
 
 // Event the main filter event
 func (f *FilterConfig) Event(ctx context.Context, event logevent.LogEvent) (logevent.LogEvent, bool) {
-	if _, ok := event.Extra[f.Key]; ok {
-		return event, false
+	if f.Overwrite == false {
+		if _, ok := event.Extra[f.Key]; ok {
```
tsaikd, line 45: `if _, ok := event.Extra[f.Key]; ok && !f.Overwrite {`

helgeolav: I have incorporated the change.

PR #165, opened 2021-09-18
Added ip2location filter

helgeolav:
ip2location is another geo database, like geoip2 - but another provider. This filter is built using their own official library and is based on the code from the geoip2 filter.

I also added some code for FileWatcher - but should this be placed somewhere else? My intention is to reuse this code other places (geoip2, ip2proxy) when the database is updated.

To be able to test this filter yourself you need to create a free account on https://lite.ip2location.com

### DIFF
--- a/filter/ip2location/filterip2location.go
+++ b/filter/ip2location/filterip2location.go
@@ -1,0 +1,235 @@
+package filterip2location
+
+import (
+	"context"
+	"github.com/fsnotify/fsnotify"
+	lru "github.com/hashicorp/golang-lru"
+	"github.com/ip2location/ip2location-go/v9"
+	"github.com/tsaikd/gogstash/config"
+	"github.com/tsaikd/gogstash/config/goglog"
+	"github.com/tsaikd/gogstash/config/logevent"
+	"net"
+	"sync"
+	"time"
+)
+
+// ModuleName is the name used in config file
+const ModuleName = "ip2location"
+
+// ErrorTag tag added to event when process ip2location failed
+const ErrorTag = "gogstash_filter_ip2location_error"
+
+// FilterConfig holds the configuration json fields and internal objects
+type FilterConfig struct {
+	config.FilterConfig
+
+	DBPath      string   `json:"db_path" yaml:"db_path"`           // ip2location .BIN file
+	IPField     string   `json:"ip_field" yaml:"ip_field"`         // IP field to get geolocation for
+	Key         string   `json:"key"`                              // destination field name
+	QuietFail   bool     `json:"quiet" yaml:"quiet"`               // fail quietly
+	SkipPrivate bool     `json:"skip_private" yaml:"skip_private"` // skip private IP addresses
+	PrivateNet  []string `json:"private_net" yaml:"private_net"`   // list of own defined private IP addresses
+	CacheSize   int      `json:"cache_size" yaml:"cache_size"`     // cache size
+
+	dbMtx sync.RWMutex    // lock for db
+	db    *ip2location.DB // database
+
+	cache        *lru.Cache
+	privateCIDRs []*net.IPNet
+
+	watcher *fsnotify.Watcher
+	ctx     context.Context
+}
+
+// DefaultCIDR is the default list of CIDRs to filter out
+var DefaultCIDR = []string{
+	"10.0.0.0/8",
+	"100.64.0.0/10",
+	"127.0.0.0/8",
+	"172.16.0.0/12",
+	"192.168.0.0/16",
+	"fc00::/7",
+	"fe80::/10",
+	"169.254.0.0/16",
+}
+
+// DefaultFilterConfig returns an FilterConfig struct with default values
+func DefaultFilterConfig() FilterConfig {
+	return FilterConfig{
+		FilterConfig: config.FilterConfig{
+			CommonConfig: config.CommonConfig{
+				Type: ModuleName,
+			},
+		},
+		Key:         ModuleName,
+		QuietFail:   false, // backwards compatible
+		SkipPrivate: false,
+		CacheSize:   100000,
+	}
+}
+
+// reloadFile reloads the file from disk invalidates the cache
+func (fc *FilterConfig) reloadFile() {
+	newDb, err := ip2location.OpenDB(fc.DBPath)
+	if err != nil {
+		goglog.Logger.Errorf("ip2location failed to update %s: %s", fc.DBPath, err.Error())
+		return
+	}
+	oldDb := fc.db
+	fc.dbMtx.Lock()
+	fc.db = newDb
+	fc.dbMtx.Unlock()
+	oldDb.Close()
+	fc.cache.Purge()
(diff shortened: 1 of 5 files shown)


tsaikd:
https://github.com/fsnotify/fsnotify
Is it what you want for file watcher?

filter/ip2location/filterip2location.go
```diff
@@ -0,0 +1,197 @@
...
+
+	var cidrs []string
+	if len(conf.PrivateNet) > 0 {
+		cidrs = conf.PrivateNet
+	} else {
+		cidrs = DefaultCIDR
+	}
+
+	for _, cidr := range cidrs {
+		_, privateCIDR, _ := net.ParseCIDR(cidr)
```
tsaikd, line 124: Should the error be ignored?

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/gomodule/redigo v2.0.0+incompatible // indirect
 	github.com/hashicorp/golang-lru v0.5.4
 	github.com/icza/dyno v0.0.0-20200205103839-49cb13720835
+	github.com/ip2location/ip2location-go v8.3.0+incompatible // indirect
```
tsaikd, line 22: There is `v9.X.X` at the repo, do you want to use the old version?

helgeolav:
Hi

Here are my answers.

1. I looked at some fsnotify alternatives, including the one you mentioned. I decided not to use any of them. The reason being that I need to add external modules that are strictly not needed. fsnotify has many open issues and PRs - and it is more work to integrate this as it over time the implementation and its API will probably change. FileWatcher does not provide real-time notifications, but that is acceptable as the underlying file changes at most once per day.
2. For the handling of error from net.ParseCIDR - it should as it can be user input. The code was copied from geoip2 and I have not changed the part of the code that was already working. (After posting this PR I also found a potential panic situation in the Event-method as the result from net.ParseIP is not validated.)
3. Not sure why ip2location 8.3 is imported, i was planning on using v9.

Let me know what way to go with filewatcher vs fsnotify. I will then address all the issues above. I will also make the changes back into geoip2 at a later point.

filter/ip2location/filewatcher.go
```diff
@@ -0,0 +1,84 @@
...
+		result = true
+		fw.fileSize = fi.Size()
+	}
+	return result
+}
+
+// runner runs in its own goroutine and checks the file on given intervals
+func (fw *FileWatcher) runner() {
+	_ = fw.hasChanged() // run once to update the internal structure before we start
+	ticker := time.NewTicker(fw.interval)
```
tsaikd, line 72:
```go
	ticker := time.NewTicker(fw.interval)
	defer ticker.Stop()
```

I think it's a better pattern for ticker.

https://pkg.go.dev/time#Ticker

filter/ip2location/filewatcher.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil, errors.New("no context passed in")
+	}
+	if callback == nil {
+		return nil, errors.New("no callback passed in")
+	}
+	fw := new(FileWatcher)
+	fw.fileName = filename
+	fw.interval = interval
+	fw.ctx = ctx
+	fw.callback = callback
```
tsaikd, line 42:
```go
fw := &FileWatcher{
foo: foo,
bar: bar,
}
```

You can init the struct with data.

filter/ip2location/filewatcher.go
```diff
@@ -0,0 +1,84 @@
...
+	if fi.Size() != fw.fileSize {
+		result = true
+		fw.fileSize = fi.Size()
+	}
+	return result
+}
+
+// runner runs in its own goroutine and checks the file on given intervals
+func (fw *FileWatcher) runner() {
+	_ = fw.hasChanged() // run once to update the internal structure before we start
```
tsaikd, line 71: do this in `NewFileWatcher()`?

`_ =` can be removed.

filter/ip2location/filewatcher.go
```diff
@@ -0,0 +1,84 @@
...
+func (fw *FileWatcher) hasChanged() bool {
+	fi, err := os.Stat(fw.fileName)
+	// if we got an error - treat it like a file does not exist
+	if err != nil {
+		fw.fileSize = -1
+		return false
+	}
+	var result bool
+	// check if modtime has changed
+	if fi.ModTime().Unix() != fw.fileTime.Unix() {
```
tsaikd, line 57: There is a `Equal()`.

https://pkg.go.dev/time#Time.Equal

filter/ip2location/filterip2location.go
```diff
@@ -0,0 +1,197 @@
...
+	IPField     string   `json:"ip_field" yaml:"ip_field"`         // IP field to get geolocation for
+	Key         string   `json:"key"`                              // destination field name
+	QuietFail   bool     `json:"quiet" yaml:"quiet"`               // fail quietly
+	SkipPrivate bool     `json:"skip_private" yaml:"skip_private"` // skip private IP addresses
+	PrivateNet  []string `json:"private_net" yaml:"private_net"`   // list of own defined private IP addresses
+	CacheSize   int      `json:"cache_size" yaml:"cache_size"`     // cache size
+
+	//initWatcher sync.Once // to make sure watcher is created only once
+	//dbMtx sync.RWMutex // lock for db
+	//db           *ip2location.DB // database
```
tsaikd, line 35: redundant lines.

tsaikd:
For the FileWatcher, my opinion is not try to make a new wheel, but it's not mandatory. It's not an easy task if you need to consider running the codes in different platforms. In your case, integrating should be simple. Just create a background go-routine at init handler function, and write the relevant callback function for detecting files changed. Anyway, it's fine to write your own tools if the other library interface is not good for you.

helgeolav:
Thanks for your feedback. It took some time to process it all. I think I got all the changes into this (updated) PR.

I moved to fsnotify as I see it is already referenced in the project. As fsnotify behaves differently on different platforms I added a timer to wait for changes to be completed. (On Windows and Mac I get more than one notification when a file is updated.)

filter/ip2location/filterip2location.go
```diff
@@ -0,0 +1,235 @@
...
+		"country_name": record.Country_long,
+	}
+	if record.City != not_supported {
+		m["city_name"] = record.City
+		m["region_name"] = record.Region
+	}
+	if record.Isp != not_supported {
+		m["ISP"] = record.Isp
+	}
+	if record.Latitude != 0 || record.Longitude != 0 {
```
tsaikd, line 217: Would it be more robust if changing `||` to `&&`?

helgeolav: If any of the numbers are non-zero then there is a coordinate where the other number is zero. It can theoretically happen.

I tried to browse through the DB5LITE CSV file to find any occurrences of this but could not find any. They claim the paid database has better accuracy (meaning more unique coordinates.) The closest entry I found to 0 was "England","Bluntisham","52.354790","0.008540".

I suggest we leave it as it is. From the CVS I only find (0,0) used on invalid IP adresses.

PR #167, opened 2021-09-25
Added ip2proxy

helgeolav:
I have added ip2proxy as a filter and also updated some of the documentation in the root folder.

### DIFF
--- a/filter/ip2proxy/filterip2proxy.go
+++ b/filter/ip2proxy/filterip2proxy.go
@@ -1,0 +1,206 @@
+package filterip2proxy
+
+import (
+	"context"
+	"github.com/fsnotify/fsnotify"
+	"github.com/ip2location/ip2proxy-go"
+	"net"
+	"sync"
+	"time"
+
+	lru "github.com/hashicorp/golang-lru"
+	"github.com/tsaikd/gogstash/config"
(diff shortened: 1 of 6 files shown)


filter/ip2proxy/filterip2proxy.go
```diff
@@ -0,0 +1,206 @@
...
+	// single-thread here
+	if c, ok := f.cache.Get(ipstr); ok {
+		record = c.(map[string]string)
+	} else {
+		var err error
+		f.dbMtx.RLock()
+		record, err = f.db.GetAll(ipstr)
+		f.dbMtx.RUnlock()
+		if err != nil {
+			if f.QuietFail {
```
tsaikd, line 139: shouldn't the logic be reversed?

helgeolav: Hi

This is also code that is copied from filtergeoip2. (And also used in filterip2location.)

I agree that it looks like it should be reversed based on the code and the comments.

How do you want to proceed? I suggest you to accept this PR, and I will then create a separate PR where I revert the logic and add the appropriate documentation for all three filters.

tsaikd: ok

PR #170, opened 2021-10-02
added disabled as an option for inputs, filters and outputs

helgeolav:
I propose to add a new parameter on all input, output and filters. This parameter - disabled - if set to true will block the configuration block to be parsed and used.

I think this parameter is useful in many ways during testing, debugging av development. For me I can easily remove a block of configuration without removing it from the configuration file.

```yaml
input:
 - type: "my-new-input"
   disabled: true
```

With a configuration like the one above I can disable the input with ease instead of changing the entire configuration file.

### DIFF
--- a/config/commonconfig.go
+++ b/config/commonconfig.go
@@ -7,7 +7,8 @@ type TypeCommonConfig interface {
 
 // CommonConfig is basic config struct
 type CommonConfig struct {
-	Type string `json:"type"`
+	Type     string `json:"type"`
+	Disabled bool   `json:"disabled" yaml:"disabled"` // if set the input/output/filter will be disabled
 }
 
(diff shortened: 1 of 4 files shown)


tsaikd:
Why not using YAML comments to temporarily disable some configuration?

```yaml
input:
# - type: "foo"
#   configure: XXX
  - type: "bar"
```

helgeolav:
With YAML you would use it more like this(my opinion):
```yaml
input:
 - type: "foo"
#   disabled: true
```

A block of configuration usually consists of more that a few lines, and commenting in/out all of them can work with YAML but it is more convenient to only comment one line than an entire section.

My personal preference is JSON over YAML and then there is no way to do do comments.

Issue #135, opened 2020-06-19
Back pressure handling document 


JoHuang:
Hi

I would like to know about if output is not available (ex : elasticsearch full disk),
what will happen?
What is the buffering mechanism and is there any parameters (like max_disk_buffer_size, max_memory_buffer_size)?

Thanks

tsaikd:
Depends on [olivere library](https://github.com/olivere/elastic) configuration, you can check the details there.

My guess is it will retry by Backoff algorithm until some special status codes.

helgeolav:
If you look at https://github.com/tsaikd/gogstash/blob/master/output/elastic/outputelastic.go you will see the default parameters buffer sizes. This is used during normal operations.

In case of issues one of three things can happen.

1. The message is not accepted and dropped.
2. The message is declined and requed. By [default](https://github.com/olivere/elastic/blob/release-branch.v7/bulk_processor.go), see line 21, it will retry messages where Elasticsearch returns with HTTP status codes 408, 429, 503, 507.
3. The message is not delivered and remain in the queue.

From what I can see what will happen is that - over time under an error condition with new messages coming in - the output will consume all memory and gogstash will crash. I can't find anywhere that a message is dropped when not delivered. The parameter ExponentialBackoffMaxTimeout does not seem to drop messages over time.

I tested with two error conditions; Elasticsearch down and Elasticsearch in read-only mode (high watermark/disk util).

Should we discuss implementing some kind of backpressure mechanism where the output can pause the input. Many inputs can support this while others have issues with it.

tsaikd:
You can have many strategies for handling delivering errors, and in my opinion, an important thing for a logging system is keeping the logs integrity. If you drop messages somewhere, you might miss a chance to be notified for the unstable system state. Case by case, sometimes you may want to drop messages to prevent memory exhaustion.

PR #175, opened 2021-10-14
Backpressure handling

helgeolav:
We discussed this a bit in #135. This is my proposal for an interface that can be used to handle backpressure.

### DIFF
--- a/config/backpressure.go
+++ b/config/backpressure.go
@@ -1,0 +1,110 @@
+package config
+
+import (
+	"github.com/tsaikd/gogstash/config/goglog"
+	"sync/atomic"
+)
+
+/*
+	To handle backpressure an output needs a mechanism to signal that it cannot process messages. And the inputs has to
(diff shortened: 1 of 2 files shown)


tsaikd:
The message channel full procedure can be handled by native go select syntax.

```go
// func (t *InputModule) Start(ctx context.Context, msgChan chan<- logevent.LogEvent) (err error)
select {
case msgChan<-msg:
  // send log event normally
default:
  // the message channel is full
}
```

tsaikd:
https://github.com/tsaikd/gogstash/commit/50d72c40aaf9147b4cc8ccac3e8dbb694640b8c6 does the interface fit your requirement?

helgeolav:
Hi

Sorry for the late answer. I have looked more into your control branch and implemented correct handling on two inputs. in #178. I have tested both inputs by modifying outputstdout (that change is not commited) to pause for a few seconds after each message received.

Would it be possible to merge my change in and commit it to master?

PR #176, opened 2021-10-21
Add GitHub action for CI

tsaikd:
@helgeolav I'm trying to fix the CI pipeline before #175, but it's blocked by the failure of output/file testing. Do you have some time to take a look?

https://github.com/tsaikd/gogstash/runs/3958503547?check_suite_focus=true

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,40 @@
+name: Gogstash
+on: [push, pull_request]
+jobs:
+  golangci:
+    strategy:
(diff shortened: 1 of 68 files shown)


helgeolav:
I can have a look tomorrow. I am not sure what to fix here - the error in the link seems to be many things; one error is that Kafka is not available. Most errors seems to be related to the configuration of the test.

Do you want me to look through the code behind each error is should I look at the build / test system and work with that to make the test work?

PR #179, opened 2021-11-22
Added backpressure to outputhttp

helgeolav:
I have worked on implementing backpressure on this output. Please review.

### DIFF
--- a/output/http/outputhttp.go
+++ b/output/http/outputhttp.go
@@ -5,6 +5,7 @@ import (
 	"container/list"
 	"context"
 	"crypto/tls"
+	"errors"
 	"io/ioutil"
 	"math/rand"
 	"net"
@@ -44,6 +45,9 @@ type OutputConfig struct {
 	IgnoreSSL           bool     `json:"ignore_ssl" yaml:"ignore_ssl"`
 	MaxQueueSize        int      `json:"max_queue_size" yaml:"max_queue_size"` // max size of queue before deleting events (-1=no limit, 0=disable)
 
+	acceptedHttpResult  map[int]struct{} // a map containing the accepted result codes
+	permanentHttpErrors map[int]struct{} // a map containing the permanent error codes
+
 	httpClient *http.Client
 	control    config.Control
 	isInPause  uint32 // set to either statusDelivering or statusPaused
@@ -60,9 +64,9 @@ func DefaultOutputConfig() OutputConfig {
 				Type: ModuleName,
 			},
 		},
-		AcceptedHttpResult:  []int{http.StatusOK, http.StatusCreated, http.StatusAccepted},
-		PermanentHttpErrors: []int{http.StatusNotImplemented, http.StatusMethodNotAllowed, http.StatusNotFound, http.StatusAlreadyReported, http.StatusHTTPVersionNotSupported},
 		RetryInterval:       30,
+		permanentHttpErrors: MapFromInts([]int{http.StatusNotImplemented, http.StatusMethodNotAllowed, http.StatusNotFound, http.StatusAlreadyReported, http.StatusHTTPVersionNotSupported}),
+		acceptedHttpResult:  MapFromInts([]int{http.StatusOK, http.StatusCreated, http.StatusAccepted}),
 		isInPause:           statusDelivering,
 	}
 }
@@ -76,6 +80,13 @@ func InitHandler(
 	conf := DefaultOutputConfig()
 	if err := config.ReflectConfig(raw, &conf); err != nil {
 		return nil, err
+	}
+
+	if len(conf.AcceptedHttpResult) > 0 {
+		conf.acceptedHttpResult = MapFromInts(conf.AcceptedHttpResult)
+	}
+	if len(conf.PermanentHttpErrors) > 0 {
+		conf.permanentHttpErrors = MapFromInts(conf.PermanentHttpErrors)
 	}
 
 	if len(conf.URLs) <= 0 {
@@ -107,8 +118,12 @@ func InitHandler(
 func (t *OutputConfig) Output(ctx context.Context, event logevent.LogEvent) (err error) {
 	// see if output has requested pause and if so just queue the event instead of trying
 	if atomic.LoadUint32(&t.isInPause) == statusPaused {
-		t.queue <- event
-		return nil
+		select {
+		case t.queue <- event:
+			return nil
+		case <-ctx.Done():
+			return errors.New("httpoutput queue context cancelled")
+		}
 	}
 	// If we are not in pause mode then call the sender method
 	return t.OutputEvent(ctx, event)
@@ -143,10 +158,10 @@ func (t *OutputConfig) OutputEvent(ctx context.Context, event logevent.LogEvent)
 		t.failedDelivery(ctx, event)
 		return err
 	}
-	if t.checkIntInDiscardedList(resp.StatusCode) {
+	if _, isinlist := t.permanentHttpErrors[resp.StatusCode]; isinlist {
 		return ErrPermanentError.New(nil, url, resp.StatusCode)
 	}
-	if !t.checkIntInAcceptedList(resp.StatusCode) {
+	if _, ok := t.acceptedHttpResult[resp.StatusCode]; !ok {
 		t.failedDelivery(ctx, event)
 		return ErrSoftError.New(nil, url, resp.StatusCode)
 	}
@@ -159,26 +174,6 @@ func (t *OutputConfig) OutputEvent(ctx context.Context, event logevent.LogEvent)
 		}
 	}
 	return nil
-}
-
-// checkIntInAcceptedList checks if code is in configured list of accepted status codes
-func (t *OutputConfig) checkIntInAcceptedList(code int) bool {
-	for _, v := range t.AcceptedHttpResult {
-		if v == code {
-			return true
-		}
-	}
-	return false
-}
-
-// checkIntInAcceptedList checks if code is in configured list of status codes where we should discard the message
-func (t *OutputConfig) checkIntInDiscardedList(code int) bool {
-	for _, v := range t.PermanentHttpErrors {
-		if v == code {
-			return true
-		}
-	}
-	return false
 }
 
 // failedDelivery receives an event that failed delivery and triggers a pause event if we have not done so already
@@ -204,7 +199,7 @@ func (t *OutputConfig) backgroundtask() {
 	for {
 		select {
 		case event := <-t.queue:
-			if (t.MaxQueueSize > 0 && t.retryqueue.Len() < t.MaxQueueSize) || t.MaxQueueSize == -1 {
+			if (t.retryqueue.Len() < t.MaxQueueSize) || t.MaxQueueSize == -1 {
 				t.retryqueue.PushBack(event)
 			}
 		case <-ticker.C:
@@ -250,4 +245,13 @@ func (t *OutputConfig) backgroundtask() {
 			}
 		}
 	}
+}
+
+// MapFromInts returns a map containing all the ints in the array
+func MapFromInts(nums []int) map[int]struct{} {
+	result := make(map[int]struct{}, len(nums))
+	for x := range nums {
+		result[x] = struct{}{}
+	}
+	return result
 }
--- a/output/http/outputhttp.go
+++ b/output/http/outputhttp.go
@@ -2,12 +2,14 @@ package outputhttp
 
 import (
 	"bytes"
+	"container/list"
 	"context"
 	"crypto/tls"
 	"io/ioutil"
 	"math/rand"
 	"net"
 	"net/http"
+	"sync/atomic"
 	"time"
 
 	"github.com/tsaikd/KDGoLib/errutil"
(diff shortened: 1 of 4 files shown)


output/http/outputhttp.go
```diff
@@ -75,12 +94,28 @@ func InitHandler(
...
 
 	return &conf, nil
 }
 
-// Output event
+// Output is receiving the event from gogstash handler
 func (t *OutputConfig) Output(ctx context.Context, event logevent.LogEvent) (err error) {
+	// see if output has requested pause and if so just queue the event instead of trying
+	if atomic.LoadUint32(&t.isInPause) == statusPaused {
+		t.queue <- event
```
tsaikd, line 110: When entered pause mode and the queue is full, `ctx` will not be able to cancel because of blocking here.

output/http/outputhttp.go
```diff
@@ -98,29 +133,121 @@ func (t *OutputConfig) Output(ctx context.Context, event logevent.LogEvent) (err
...
 		if v == code {
 			return true
 		}
 	}
 	return false
 }
+
+// checkIntInAcceptedList checks if code is in configured list of status codes where we should discard the message
+func (t *OutputConfig) checkIntInDiscardedList(code int) bool {
+	for _, v := range t.PermanentHttpErrors {
```
tsaikd, line 176: Check if value is in a map will get a better performance. The slice from config can be converted to map in init phase.

output/http/outputhttp.go
```diff
@@ -98,29 +133,121 @@ func (t *OutputConfig) Output(ctx context.Context, event logevent.LogEvent) (err
...
+func (t *OutputConfig) checkIntInAcceptedList(code int) bool {
 	for _, v := range t.AcceptedHttpResult {
 		if v == code {
 			return true
 		}
 	}
 	return false
 }
+
+// checkIntInAcceptedList checks if code is in configured list of status codes where we should discard the message
```
tsaikd, line 174: `checkIntInAcceptedList` -> `checkIntInDiscardedList`

output/http/outputhttp.go
```diff
@@ -98,29 +133,121 @@ func (t *OutputConfig) Output(ctx context.Context, event logevent.LogEvent) (err
...
+					event := e.Value.(logevent.LogEvent)
+					t.retryqueue.Remove(e)
+					go func() {
+						ctx, cancel := context.WithTimeout(context.Background(), dur)
+						err := t.OutputEvent(ctx, event)
+						if err != nil {
+							goglog.Logger.Error("outputhttp background sendone ", err.Error())
+						}
+						cancel()
+					}()
```
tsaikd, line 224: Why using another goroutine to retry?

output/http/outputhttp.go
```diff
@@ -98,29 +133,121 @@ func (t *OutputConfig) Output(ctx context.Context, event logevent.LogEvent) (err
...
+
+// backgroundtask is running in the background and adds new events to the queue, tries to send them out on the RetryInterval.
+func (t *OutputConfig) backgroundtask() {
+	dur := time.Duration(t.RetryInterval) * time.Second
+	ticker := time.NewTicker(dur)
+	defer ticker.Stop()
+	for {
+		select {
+		case event := <-t.queue:
+			if (t.MaxQueueSize > 0 && t.retryqueue.Len() < t.MaxQueueSize) || t.MaxQueueSize == -1 {
```
tsaikd, line 207: `t.MaxQueueSize > 0` is not required.

output/http/outputhttp.go
helgeolav: I have done some testing new using the bench. I have tested three different variants based on the defaults in PermanentHttpErrors. The three tests are:

1. The already implemented code.
2. A map
3. A minor improvement on the implemented code.

```
goos: darwin
goarch: amd64
pkg: github.com/tsaikd/gogstash/output/http
cpu: Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
BenchmarkCheckIntInAcceptedList-8   	193956784	         6.089 ns/op
BenchmarkMap-8                      	142291500	         8.596 ns/op
BenchmarkSliceIdx-8                 	220563901	         5.428 ns/op
PASS
ok  	github.com/tsaikd/gogstash/output/http	5.652s
```

So based on these findings running on my Mac, I will rewrite the code to implement the index-method instead. In some cases I think this can be improved even more by sorting the list and exiting if the number is higher. But I don't think it is worth the work as it performs very good already.

The map was implemented in this way:
```golang
var result bool
// benchMap is a test where a map is tested instead of a list that is passed
type benchMap struct {
	list map[int]struct{}
}

func (b *benchMap) IsinList(number int) bool {
	_, ok := b.list[number]
	return ok
}

func newBenchMap(nums ...int) benchMap {
	result := benchMap{}
	result.list = make(map[int]struct{})
	for x := range nums {
		result.list[nums[x]] = struct{}{}
	}
	return result
}

func BenchmarkMap(b *testing.B) {
	var r bool
	o := newBenchMap(http.StatusNotImplemented, http.StatusMethodNotAllowed, http.StatusNotFound, http.StatusAlreadyReported, http.StatusHTTPVersionNotSupported)
	const wantresult = 200
	for n:= 0; n < b.N; n++ {
		r = o.IsinList(wantresult)
	}
	result = r
}
```

output/http/outputhttp.go
helgeolav: Is is true that the code will block if the queue gets full. I am not sure the best approach here. I have tried to write the quene handler as fast as possible to minimize this problem.

In case there is congestion this is perhaps the best outcome anyway? The size of the queue channel could be adjusted to the size of the chsize (but that property is not accessible for the output?) or be configured so the user of gogstash can choose for themselves?

Please advice.

output/http/outputhttp.go
helgeolav: The price of copy & paste. Fixed.

output/http/outputhttp.go
helgeolav: When I send something back to OutputEvent the call can block for a long time in case there are timeout issues present -  this will pause the entire background task until OutputEvent has returned. A goroutine is used to mitigate this.

As already mentioned this is critical code because I need to be able to process new events coming into the queue.

output/http/outputhttp.go
helgeolav: Fixed.

helgeolav:
I have pushed out all the changes we talked about above. I have not done anything with the queue-issue in Output() - I need more input.

output/http/outputhttp.go
tsaikd: Your benchmark codes are not really benchmark the core function. Here is my benchmark samples in the old fashion MacBook Air.

```go
type benchMap struct {
	data map[int]struct{}
}

func (b *benchMap) In(number int) bool {
	_, ok := b.data[number]
	return ok
}

func newBenchMap(nums ...int) benchMap {
	result := benchMap{
		data: map[int]struct{}{},
	}
	for _, code := range nums {
		result.data[code] = struct{}{}
	}
	return result
}

type benchSlice struct {
	slice []int
}

func (b *benchSlice) ByIndex(number int) bool {
	for idx := range b.slice {
		if b.slice[idx] == number {
			return true
		}
	}
	return false
}

func (b *benchSlice) ByValue(number int) bool {
	for _, v := range b.slice {
		if v == number {
			return true
		}
	}
	return false
}

func newBenchSlice(nums ...int) benchSlice {
	return benchSlice{nums}
}

func BenchmarkMap(b *testing.B) {
	nums := []int{
		http.StatusNotImplemented,
		http.StatusMethodNotAllowed,
		http.StatusNotFound,
		http.StatusAlreadyReported,
		http.StatusHTTPVersionNotSupported,
	}
	bSlice := newBenchSlice(nums...)
	bMap := newBenchMap(nums...)

	// do some unit tests to make sure the functions work
	require.False(b, bSlice.ByIndex(http.StatusOK))
	require.False(b, bSlice.ByValue(http.StatusOK))
	require.False(b, bMap.In(http.StatusOK))
	require.True(b, bSlice.ByIndex(http.StatusMethodNotAllowed))
	require.True(b, bSlice.ByValue(http.StatusMethodNotAllowed))
	require.True(b, bMap.In(http.StatusMethodNotAllowed))

	b.Run("slice ByIndex", func(b *testing.B) {
		for n := 0; n < b.N; n++ {
			bSlice.ByIndex(http.StatusOK)
		}
	})
	b.Run("slice ByValue", func(b *testing.B) {
		for n := 0; n < b.N; n++ {
			bSlice.ByValue(http.StatusOK)
		}
	})
	b.Run("map", func(b *testing.B) {
		for n := 0; n < b.N; n++ {
			bMap.In(http.StatusOK)
		}
	})

	// generate random testing numbers
	const randomCount = 1_000_000
	randomStatus := make([]int, randomCount)
	random := rand.New(rand.NewSource(time.Now().UnixMicro()))
	for i := 0; i < randomCount; i++ {
		num := random.Intn(100)
		if num > 80 { // 80% cases should be OK
			randomStatus[i] = http.StatusOK
		} else {
			randomStatus[i] = nums[num%len(nums)]
		}
	}

	b.Run("random slice ByIndex", func(b *testing.B) {
		for n := 0; n < b.N; n++ {
			bSlice.ByIndex(randomStatus[n%randomCount])
		}
	})
	b.Run("random slice ByValue", func(b *testing.B) {
		for n := 0; n < b.N; n++ {
			bSlice.ByValue(randomStatus[n%randomCount])
		}
	})
	b.Run("random map", func(b *testing.B) {
		for n := 0; n < b.N; n++ {
			bMap.In(randomStatus[n%randomCount])
		}
	})
}
```

```
goos: darwin
goarch: amd64
pkg: github.com/tsaikd/gogstash/output/http
cpu: Intel(R) Core(TM) i5-5350U CPU @ 1.80GHz
BenchmarkMap
BenchmarkMap/slice_ByIndex
BenchmarkMap/slice_ByIndex-4         	 6798596	       193.0 ns/op	       0 B/op	       0 allocs/op
BenchmarkMap/slice_ByValue
BenchmarkMap/slice_ByValue-4         	12674724	       111.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkMap/map
BenchmarkMap/map-4                   	25927770	        59.93 ns/op	       0 B/op	       0 allocs/op
BenchmarkMap/random_slice_ByIndex
BenchmarkMap/random_slice_ByIndex-4
 8445800	       159.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkMap/random_slice_ByValue
BenchmarkMap/random_slice_ByValue-4  	 8104420	       149.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkMap/random_map
BenchmarkMap/random_map-4            	16909435	        86.29 ns/op	       0 B/op	       0 allocs/op
```

output/http/outputhttp.go
tsaikd: There is a non-blocking channel pattern in go.

```go
select {
case <- ctx.Done():
	// parent context canceled
case <- channel:
	// process the channel data
default:
	// if the channel is full
}
```

https://stackoverflow.com/questions/52587537/non-blocking-channel-in-go

helgeolav: Hi

I don't see any other outcome from this flow than just wait for the event to be put into the queue. So if the code blocks then I think that is ok. The time it takes to queue the event should still be shorter than the normal wait-to-process-an-event.

The output handler (output.go / startOutputs()) blocks further processing until the message is queued or processed. I think that is acceptable.

Please advice.

output/http/outputhttp.go
helgeolav: Hi

I think the fundamental difference is that you are testing with lots of elements where maps really shine. In practice there should be less than 10 elements to compare against (not that many HTTP error codes) and then it is faster to iterate over the array. I have found several articles about this issue. One is [here](https://boltandnuts.wordpress.com/2017/11/20/go-slice-vs-maps/).

If you want I will just rewrite the code to use maps instead of iterators.

For completeness, below you have my benchmarking file. I have not tried to run the tests with lots of elements as I tried to mimic how I expect the code to be used. I have no experience with benchmarking and everything is done using code I found online.

```go
package outputhttp

// run with go test -bench=. in this directory

import (
	"net/http"
	"testing"
)

var result bool

// BenchmarkCheckIntInAcceptedList checks the applied method
func BenchmarkCheckIntInAcceptedList(b *testing.B) {
	o := DefaultOutputConfig()
	var r bool
	const wantresult = 200
	for n := 0; n < b.N; n++ {
		r = o.checkIntInDiscardedList(wantresult)
	}
	result = r
}

// benchMap is a test where a map is tested instead of a list that is passed
type benchMap struct {
	list map[int]struct{}
}

func (b *benchMap) IsinList(number int) bool {
	_, ok := b.list[number]
	return ok
}

func newBenchMap(nums ...int) benchMap {
	result := benchMap{}
	result.list = make(map[int]struct{})
	for x := range nums {
		result.list[nums[x]] = struct{}{}
	}
	return result
}

func BenchmarkMap(b *testing.B) {
	var r bool
	o := newBenchMap(http.StatusNotImplemented, http.StatusMethodNotAllowed, http.StatusNotFound, http.StatusAlreadyReported, http.StatusHTTPVersionNotSupported)
	const wantresult = 200
	for n := 0; n < b.N; n++ {
		r = o.IsinList(wantresult)
	}
	result = r
}

// sliceIdx is a test where we check by index, not by range where the values are copied over
type sliceIdx struct {
	list []int
}

func newSliceIdx(nums ...int) sliceIdx {
	return sliceIdx{list: nums}
}

func (b *sliceIdx) IsinList(number int) bool {
	for x := range b.list {
		if b.list[x] == number {
			return true
		}
	}
	return false
}

func BenchmarkSliceIdx(b *testing.B) {
	o := newSliceIdx(http.StatusNotImplemented, http.StatusMethodNotAllowed, http.StatusNotFound, http.StatusAlreadyReported, http.StatusHTTPVersionNotSupported)
	var r bool
	const wantresult = 200
	for n := 0; n < b.N; n++ {
		r = o.IsinList(wantresult)
	}
	result = r
}
```

output/http/outputhttp.go
tsaikd: It might be ok in this case because the cooperated functions take care of the blocking issue. (a background go routine consumes the queue)

But I don't think it's a good practice. There is a `ctx` in the parameter, so, at least, you should take care of the parent context canceled case instead of blocking the go routine.

output/http/outputhttp.go
tsaikd: In my codes, `bSlice` and `bMap` are both initialized by the same input `nums` which only has 5 elements like yours.

In the first 3 benchmarks, I test the worst case for slice, but it's the most case in real life.

In the last 3 benchmarks, I simulate if there are 20% errors.

In your codes, there are 2 points should be considering.

1. Exclude initialization costs.
2. Exclude non-core function codes.

```diff
func BenchmarkMap(b *testing.B) {
-	var r bool
	o := newBenchMap(http.StatusNotImplemented, http.StatusMethodNotAllowed, http.StatusNotFound, http.StatusAlreadyReported, http.StatusHTTPVersionNotSupported)
	const wantresult = 200

+	b.ResetTimer()
	for n := 0; n < b.N; n++ {
-		r = o.IsinList(wantresult)
+		o.IsinList(wantresult)
	}
-	result = r
}
```

More examples

```go
func BenchmarkSleep1(b *testing.B) {
	for i := 0; i < b.N; i++ {
		time.Sleep(10 * time.Millisecond)
	}
}

func BenchmarkSleep2(b *testing.B) {
	time.Sleep(1000 * time.Millisecond)

	for i := 0; i < b.N; i++ {
		time.Sleep(10 * time.Millisecond)
	}
}

func BenchmarkSleep3(b *testing.B) {
	time.Sleep(1000 * time.Millisecond)

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		time.Sleep(10 * time.Millisecond)
	}
}
```

```
goos: darwin
goarch: amd64
pkg: github.com/tsaikd/gogstash/output/http
cpu: Intel(R) Core(TM) i5-5350U CPU @ 1.80GHz
BenchmarkSleep1-4   	     100	  10986617 ns/op
BenchmarkSleep2-4   	       1	1011198468 ns/op
BenchmarkSleep3-4   	     100	  11316856 ns/op
PASS
```

Your referral https://boltandnuts.wordpress.com/2017/11/20/go-slice-vs-maps/ might not be a good sample. Can `b.N` be changed in code? I'm not sure. But I think the correct way to limit the benchmark round is by passing argument, e.g. `go test -benchtime=100x` means `b.N = 100`

output/http/outputhttp.go
helgeolav: I have rewritten the code to handle context now.

output/http/outputhttp.go
helgeolav: I rewrote the code to use maps.

PR #180, opened 2021-12-11
Refactored queue

helgeolav:
After looking into outputhttp I found lots of code that I can reuse. Instead of writing all code over again I propose to move much of the backpressure/queueing logic out to a separate class that can be reused several places. I hope this can make it easier to implement backpressure in more outputs and still have readable code.

### DIFF
--- a/config/queue/simplequeue.go
+++ b/config/queue/simplequeue.go
@@ -1,0 +1,156 @@
+package queue
+
+import (
+	"container/list"
+	"context"
+	"github.com/tsaikd/gogstash/config"
+	"github.com/tsaikd/gogstash/config/goglog"
+	"github.com/tsaikd/gogstash/config/logevent"
+	"sync/atomic"
+	"time"
+)
+
+// simpleQueue is an implementation of QueueReceiver with an easy retry-logic
+type simpleQueue struct {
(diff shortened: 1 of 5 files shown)


config/queue/simplequeue.go
```diff
@@ -0,0 +1,153 @@
...
+		case <-t.ctx.Done():
+			return ErrContextCancelled.New(nil)
+		}
+	}
+	// If we are not in pause mode then call the sender method
+	return t.output.OutputEvent(ctx, event)
+}
+
+// backgroundtask is running in the background and adds new events to the queue, tries to send them out on the RetryInterval.
+func (t *simpleQueue) backgroundtask() {
```
tsaikd, line 97: https://github.com/tsaikd/gogstash/runs/4495112091

`backgroundtask()` is not triggered.

helgeolav: I have solved the issue and added a test case.

tsaikd: https://github.com/tsaikd/gogstash/runs/4509922290

additional lint issues

helgeolav: I looked into the lint. It seems to be in the test file and because I did not handle an error that was not important for the test. I have posted an update on this.

I did:

funcWithError()

now changed to:

_ = funcWithError()

PR #182, opened 2021-12-23
Added backpressure to outputnsq

helgeolav:
I have added backpressure to outputnsq. This change required some changes to simplequeue as I also needed to handle events that was sent through an codec.

PR #185, opened 2022-01-26
Implement gelf output

mderasse:
That implementation allow direct output to Graylog infrastructure.

Multiple implementation of gelf_writer can be found over github.
Mostly the one part of that PR is a classical version with a few adaptation to allow configuration of compression / chunkSize, ...

### DIFF
--- a/output/gelf/gelf_writer.go
+++ b/output/gelf/gelf_writer.go
@@ -1,0 +1,462 @@
+// Copyright 2012 SocialCode. All rights reserved.
+// Use of this source code is governed by the MIT
+// license that can be found in the LICENSE file.
+
+package outputgelf
+
+import (
+	"bytes"
+	"compress/flate"
+	"compress/gzip"
+	"compress/zlib"
+	"crypto/rand"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"strings"
+	"sync"
+	"time"
+)
+
+// regex used later
+var forbiddenExtraKeys = [7]string{
+	"_version",
+	"_host",
+	"_short_message",
+	"_full_message",
+	"_timestamp",
+	"_level",
+}
+
+type GELFConfig struct {
+	Host             string
+	ChunkSize        int          // Default Value: 1420
+	CompressionLevel int          // Default: 1
+	CompressionType  CompressType // Default: Gzip
+}
+
+type GELFWriter interface {
+	WriteCustomMessage(m *Message) error
+	WriteMessage(sm *SimpleMessage) error
+}
+
+// UDPWriter implements io.Writer and is used to send both discrete
+// messages to a graylog2 server, or data from a stream-oriented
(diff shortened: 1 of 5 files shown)


tsaikd:
Please check https://github.com/tsaikd/gogstash/runs/4974337609?check_suite_focus=true for lint issue.

helgeolav:
Will it be possible to add some kind of backpressure control on this output? This is a way to ask the input to stop receiving data in case of delivery issues. Look at #182 where I implemented this for another output. I can also help you with this if it is not clear how to do it.

Here is a guide on how to do it: https://github.com/tsaikd/gogstash/tree/master/config/queue

mderasse:
Thank you for your review and comment

I will take a look a those this weekend,
I should also probably update the ReadMe and the documentation regarding the configuration of the GELF output.

mderasse:
Lint issue fixed (sorry about that ..., i should have check)
Queue implemented, one potential issue with the queue is that if one of the host fail (and not the 1st one of the list), we might resend the message.

tsaikd:
The interface of the queue was changed, so you need to fix for it.

output/gelf/gelf_writer.go
```diff
@@ -0,0 +1,461 @@
...
+func constructMessage(sm *SimpleMessage) *Message {
+
+	message := Message{
+		Extra:     sm.Extra,
+		Level:     6, // Default: Info
+		Timestamp: int64(time.Now().UnixNano() / int64(time.Millisecond) / 1000),
+		Version:   "1.1",
+	}
+
+	message.Host, _ = os.Hostname()
```
tsaikd, line 181: You can panic on this err to keep the returning type simple, Assume os.Hostname() should not return err.

One more thing, the hostname can be cached if you don't want to support the hostname changes in the process life cycle.

output/gelf/outputgelf.go
```diff
@@ -0,0 +1,138 @@
...
+
+	return conf.queue, nil
+}
+
+// OutputEvent handle message from the queue
+func (t *OutputConfig) OutputEvent(ctx context.Context, event logevent.LogEvent) (err error) {
+	var host string
+	var level int32
+	for k, v := range event.Extra {
+		if strings.ToLower(k) == "host" || strings.ToLower(k) == "hostname" {
```
tsaikd, line 99: How about saving `strings.ToLower(k)` to a variable? You use it many times in the block.

mderasse: Since I'm handling multiple potential "Extra" (like host & hostname) I think it will be better with a if condition.

output/gelf/gelf_writer.go
mderasse: os.Hostname was already previously called on the creation of the Writer.
I did a global var called hostname and that hostname is fill on the creation of the Writer (And it panic if there is an error)

mderasse:
Thank you :)

Issue #211, opened 2023-04-29
Ctrl+C does not kill worker processes


Mizaro:
While trying the project on my local machine. I did the following:
```terminal
go run .
ps aux | grep gog
```
And noticed that the worker processes are still alive.
```
user1    5896 21.6  0.3 1628392 27444 pts/6   Sl   11:00   0:01 /tmp/go-build3520306496/b001/exe/gogstash worker
user1    5897 21.0  0.3 1628136 31328 pts/6   Sl   11:00   0:01 /tmp/go-build3520306496/b001/exe/gogstash worker
```

My config is `config.yaml` is:
```yaml
chsize: 1000
worker: 2
event:
  sort_map_keys: false
  remove_field: ['@metadata']

input:
  - type: lorem

output:
  - type: stdout
    codec: json
```

Mizaro:
Closing because it is solved by #212

Issue #199, opened 2022-11-17
Cannot stop program


KnifeFly:
i clone and build, and start program: 
```
./gogstash worker --config ./gogstash.yml
```

gogstash.yml config file content:
```
worker: 2
chsize: 1000

input:
  - type: file
    path: /tmp/logstash-tutorial.log
    start_position: beginning
    sincedb_write_interval: 5

output:
  - type: stdout
    codec: json
```

while i want to stop the program, so i enter `command + c` to interrupt, but seems the program cannot stop, just print `interrupt` information, i sam not sure that happen.

<img width="750" alt="image" src="https://user-images.githubusercontent.com/13764181/202457227-9b6dc137-232c-43bd-957e-78ccd8dc1592.png">

```
util.go:116 [info] interrupt
```

machine: macos 12.5.1
go version: go version go1.18.5 darwin/amd64

tsaikd:
Is your input file huge for each line?

KnifeFly:
@tsaikd hihi, i am I'm afraid not, only 4 lines for /tmp/logstash-tutorial.log

tsaikd:
Unfortunately, I cannot reproduce your state with your input file.
```
➜  gogstash git:(master) ✗ uname -mprsv
Darwin 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:19:52 PDT 2022; root:xnu-8020.140.49~2/RELEASE_ARM64_T6000 arm64 arm
➜  gogstash git:(master) ✗ go version
go version go1.19.3 darwin/arm64
➜  gogstash git:(master) ✗ git show-ref HEAD
0813dbd086f870631c5dd25d628002118b52ffa7 refs/remotes/origin/HEAD
➜  gogstash git:(master) ✗ ./gogstash worker --config ./gogstash.yml
2022/11/22 22:49:28 gogstash.go:61 [info] gogstash started...
[...]

PR #212, opened 2023-04-29
Fixed deadlock on codec when exiting program

helgeolav:
I have been looking at #211 and #199, and it seems to be related to that JSON Encode() does not check the context and in some cases ends up blocking as dataChan is full. At least on my tests this seems to solve the problem.

### DIFF
--- a/codec/json/codecjson.go
+++ b/codec/json/codecjson.go
@@ -102,9 +102,8 @@ func (c *Codec) Encode(ctx context.Context, event logevent.LogEvent, dataChan ch
 	}
 	select {
 	case <-ctx.Done():
-		return
-	default:
-		dataChan <- output
+		return false, nil
+	case dataChan <- output:
 	}
 	return true, nil
 }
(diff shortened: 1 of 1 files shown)


Mizaro:
Thanks for trying to solve it. I applied your code changes to my code, and the problem still needs to be fixed.

At first, I saw that all of the workers were always alive. Now sometimes they die, and sometimes they don't.

codec/json/codecjson.go
```diff
@@ -95,12 +95,17 @@ func (c *Codec) DecodeEvent(data []byte, event *logevent.LogEvent) (err error) {
...
 	if err != nil {
 		return false, err
 	}
-	dataChan <- output
+	select {
+	case <-ctx.Done():
+		return
+	default:
+		dataChan <- output
+	}
```
tsaikd, line 108:
```go
select {
case <-ctx.Done():
  return
case dataChan <- output:
}
```

Mizaro: Nice, this fix the problem

PR #215, opened 2023-04-29
deadlock in azure codec

helgeolav:
Fixing the same issue as in #212 for the Azure codec as well.


Mizaro (approved):
Maybe we can use a common function?

helgeolav:
> Maybe we can use a common function?

That would be helpful. In many ways much of the code is boilerplate (copy/paste). Like the codecs. If you look at the code for the filter ip2location, ip2proxy and geoip2 - they are almost identical. The only difference is what database they open and how it is accessed(each have its own API).

But the best way to go around this? Composition, interfaces or both? The reason I mention these three filters is because I plan to upgrade hashicorp-lru for them, and it the same changes three places. (Moving from interface to generics.)

tsaikd:
Please rebase the PR for merging

helgeolav:
> Please rebase the PR for merging

rebase is not my friend. Do you have q quick command line on what I have to do? I think my quickest route is to just create a new PR.

tsaikd:
Because you merged other commits which are not in the master branch, so there are no one line command to rebase quickly.

1. git checkout -b PR-215 aed1915bdf5f0c6de3afe97df230c269cd000770

checkout the master branch

2. git cherry-pick 7fc75a2a4a66eacaadd010a626fae169ade64688

pick up the commit

3. git push -f <REMOTE> PR-215:staleworker

REMOTE might be `origin`

tsaikd:
merged

PR #216, opened 2023-04-29
Testing Interrupt Signal

Mizaro:
Added Test for code in #212 and had to fix something to make it work.


Mizaro:
@helgeolav Take a look, I added test for #212

cmd/worker_windows.go
```diff
@@ -78,5 +79,15 @@ func startWorkers(ctx context.Context, workerNum int) error {
...
 		return waitWorkers(ctx, pids, handles, args, attr)
 	})
-	return waitSignals(ctx)
+
+	signaled, err := waitSignals(ctx)
+	if signaled {
+		for _, pid := range pids {
+			p, err := os.FindProcess(pid)
+			if err == nil {
+				_ = p.Signal(os.Interrupt)
```
tsaikd, line 88: https://github.com/golang/go/blob/go1.20.3/src/os/exec.go#L135

main_test.go
```diff
@@ -0,0 +1,88 @@
...
+	if err := cmd.Start(); err != nil {
+		t.Fatalf("Failed to start the program: %v", err)
+	}
+
+	// Get the process ID of the main process
+	mainPID := int32(cmd.Process.Pid)
+
+	// Find all child processes of the main process
+	time.Sleep(waitingTime) // Wait for the program to spawn child processes
+	childProcesses, err := getChildProcesses(mainPID)
```
tsaikd, line 39:
```go
func waitChildProcesses(parentPID int32, timeout time.Duration) ([]int32, error) {
	ctx, cancel := context.WithTimeout(context.Background(), timeout)
	defer cancel()

	interval := 100 * time.Millisecond
	timer := time.NewTimer(interval)
	defer timer.Stop()

	for {
		select {
		case <-ctx.Done():
			return nil, context.DeadlineExceeded
		case <-timer.C:
			childPIDs, err := getChildProcesses(parentPID)
			if err != nil {
				return nil, err
			}
			if len(childPIDs) > 0 {
				return childPIDs, nil
			}
			timer.Reset(interval)
		}
	}
}
```

main_test.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+const ExecutableNamePrefix = "gogstash_test"
+const waitingTime = 100 * time.Millisecond
+
+
+func TestProgramTermination(t *testing.T) {
+	// Start the main program in a separate process
+  executablePath := fmt.Sprintf("./%s_%s", ExecutableNamePrefix, randomstring.RandString(6))
+	_, _ = exec.Command("go", "build", "-o", executablePath).CombinedOutput()
```
tsaikd, line 24: Ignore error?

main_test.go
```diff
@@ -0,0 +1,88 @@
...
+	// Find all child processes of the main process
+	time.Sleep(waitingTime) // Wait for the program to spawn child processes
+	childProcesses, err := getChildProcesses(mainPID)
+	t.Logf("Children are %s", strings.Join(lo.Map(childProcesses, func(v int32, _ int) string { return strconv.Itoa(int(v)) }), ", "))
+	if err != nil {
+		t.Fatalf("Failed to get child processes: %v", err)
+	}
+	if len(childProcesses) < 1 {
+		t.Fatal("Cannot test if there are no child processes")
+	}
```
tsaikd, line 46:
```go
	// Cleanup created child processes
	defer func() {
		for _, pid := range childProcesses {
			cp, err := process.NewProcess(pid)
			if err == nil {
				err = cp.Kill()
				if err != nil {
					t.Errorf("Cannot terminate process with PID %d", pid)
				}
			}
		}
	}()
```

main_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// Recalling `Processes` to get a fresh information from the OS. Important to force `process.NewProcess` to return valid information.
+	_, err = process.Processes()
+	if err != nil {
+		t.Fatalf("Failed to update processes information: %v", err)
+	}
+
+	// Check whether all child processes have terminated. 
+	for _, pid := range childProcesses {
+		cp, err := process.NewProcess(pid)
```
tsaikd, line 64:
```go
		exist, err := process.PidExists(pid)
		if err != nil {
			t.Fatal(err)
		}
		if exist {
			t.Errorf("Process with PID %d did not terminate.", pid)
		}
```

main_test.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func getChildProcesses(parentPID int32) ([]int32, error) {
+	var childPIDs []int32
+	allProcesses, err := process.Processes()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get all processes: %v", err)
+	}
+	for _, p := range allProcesses {
+		ppid, _ := p.Ppid()
```
tsaikd, line 82: Ignore error?

cmd/worker.go
```diff
@@ -9,14 +9,16 @@ import (
...
 
 	select {
 	case <-ctx.Done():
 	case sig := <-osSignalChan:
+		signaled = true
 		goglog.Logger.Info(sig)
 	}
-	return nil
+	return signaled, nil
 }
```
tsaikd, line 24:
```go

func waitSignals(ctx context.Context) (os.Signal, error) {
	osSignalChan := make(chan os.Signal, 1)
	signal.Notify(osSignalChan, os.Interrupt, syscall.SIGTERM)

	select {
	case <-ctx.Done():
		return nil, nil
	case sig := <-osSignalChan:
		goglog.Logger.Info(sig)
		return sig, nil
	}
}
```

main_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/shirou/gopsutil/v3/process"
+)
+
+const ExecutableNamePrefix = "gogstash_test"
+const waitingTime = 100 * time.Millisecond
+
+
+func TestProgramTermination(t *testing.T) {
+	// Start the main program in a separate process
+  executablePath := fmt.Sprintf("./%s_%s", ExecutableNamePrefix, randomstring.RandString(6))
```
tsaikd, line 23:
```go
	randBytes := make([]byte, 4)
	if _, err := rand.Read(randBytes); err != nil {
		t.Fatal(err)
	}
	executablePath := fmt.Sprintf("./%s_%x", ExecutableNamePrefix, randBytes)
```

main_test.go
Mizaro: Nice! I will add it! :)

main_test.go
Mizaro: I will convert to t.Fatal

main_test.go
Mizaro: If I could not the ppid of a process it is just meant that this is not the process I am looking for :)

cmd/worker_windows.go
Mizaro: I used os.Kill here, assuming we will get to the same result :/

Mizaro:
@tsaikd I fixed and commented on all of your comments :)

tsaikd:
the package randomstring could be removed?

main_test.go
tsaikd: It's better to have the explicit error checking, and add some comments if it's reasonable. e.g.
```go
for _, p := range allProcesses {
  ppid, err := p.Ppid()
  if err != nil {
    // blahblah...
    continue
  }
}
```

cmd/worker.go
```diff
@@ -9,14 +9,15 @@ import (
 	"github.com/tsaikd/gogstash/config/goglog"
 )
 
-func waitSignals(ctx context.Context) error {
+func waitSignals(ctx context.Context) (os.Signal, error) {
```
tsaikd, line 12: returned error is useless?

Mizaro: pre-commit would have told me to remove it 😇

main_test.go
tsaikd: any update?

tsaikd:
lint error

https://github.com/tsaikd/gogstash/actions/runs/4857667340/jobs/8658359480?pr=216#step:3:80

Mizaro:
@tsaikd fixed + squashed

main_test.go
Mizaro: fixed

Mizaro:
@tsaikd thanks

PR #217, opened 2023-04-29
Added Pre-commit and Makefile

Mizaro:
Added pre-commit configuration, updated golangci-lint to use more tools, and applied most of the changes to the existing code.

All commented-out linters in golangci-lint that I think should be part of the project but there are too many changes already in this PR so I postponed adding those to another PR.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,88 @@
+linters-settings:
+  dupl:
+    threshold: 100
+  goconst:
+    min-len: 2
+    min-occurrences: 1
+  gocritic:
+    enabled-tags:
+      - diagnostic
+      - experimental
+      - opinionated
+      - performance
+      - style
+    disabled-checks:
+      - dupImport # https://github.com/go-critic/go-critic/issues/845
+      - ifElseChain
+  gocyclo:
+    min-complexity: 30 # Recommended between 10-20
+  gofmt:
+    rewrite-rules:
+      - pattern: 'interface{}'
+        replacement: 'any'
+  goimports:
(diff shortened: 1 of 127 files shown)


tsaikd:
I think the `Makefile` and `.pre-commit-config.yaml` should move out of the PR, because they are not the standard for golang project and keep the code base clean.

Mizaro:
> I think the `Makefile` and `.pre-commit-config.yaml` should move out of the PR, because they are not the standard for golang project and keep the code base clean.

All of those changes are made by `golangci-lint` new config and the `pre-commit run --all-files` command. Let's enforce those rules in pre-commit as well as in the CI, and to do it, I would expect contributors to set up their machine to use  `.pre-commit-config.yaml`.

I though that instead of assuming people know how to set up the pre-commit, I would create a Makefile for them.
Do you prefer a CONTRIBUTION.md file that explains how to set up the pre-commit instead of the Makefile and leaving the pre-commit file?

tsaikd:
https://github.com/tsaikd/gogstash/blob/master/.github/workflows/build.yml
You can add the checks/lints in Github Actions

Mizaro:
> https://github.com/tsaikd/gogstash/blob/master/.github/workflows/build.yml You can add the checks/lints in Github Actions

I added `.golangci.yaml` config as well, so the CI will check stuff.

I will delete Makefile, but I would like to leave the .pre-commit-config file. WDYT?

tsaikd:
Please squash the commits before merging.

Mizaro:
@tsaikd Thanks for approving! I don't have write access to merge it, Am I missing something? Do you think I should merge it to my Fork first?

PR #218, opened 2023-05-03
Enable bodyclose in golangci-lint

Mizaro:
Enable bodyclose check in golangci-lint and fix other code in the project

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -53,7 +53,7 @@ linters-settings:
 linters:
   disable-all: true
   enable:
-      #- bodyclose
+    - bodyclose
     - dogsled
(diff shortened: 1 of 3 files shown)


Mizaro:
> https://github.com/timakin/bodyclose
> 
> * Pattern
> 
> ```go
> resp, err := http.Get("http://example.com/")
> if err != nil {
> 	// handle error
> }
> defer resp.Body.Close() // OK
> ```

I remember that is way is not enough. Something about 3XX or 4XX responses. I found the following 
https://www.ziccardi.net/blog/tips-and-tricks/the-good-way-to-close-a-response-body but I see there is no explanation

PR #237, opened 2024-06-12
Implemented Sentry

cdmx1:
This PR integrates Sentry into Gogstash, to enhance error monitoring and logging capabilities. 
With this integration, we can now capture and report errors more effectively, including those that occur during requests to Elastic/OpenSearch.

Sentry can be enabled by setting the GS_SENTRY_DSN environment variable.
This can be done in export/linux environments or within Docker environments.


tsaikd:
How about implementing in the application level logging tool (goglog)?

cdmx1:
> How about implementing in the application level logging tool (goglog)?
Hi @tsaikd,
But Sentry provides real-time error tracking and alerts, ensuring that issues are identified and addressed promptly with email notifications, frequency of alert and its relevance to previous occurrences..
Also sentry consolidates error logs across various environments into a single platform, making it easier to manage and analyse errors.
sentry also captures extensive contextual information about errors, such as stack traces, user actions, and environment details, which greatly aids in debugging and understanding the root cause of issues.

Sentry can work alongside goglog, complementing the existing logging mechanism without introducing significant overhead.

tsaikd:
In the PR, you only send the the error logs to sentry in output/elastic module. I mean, you can send the application level error logs to sentry in goglog package which will imply for wider error logs.

cdmx1:
> In the PR, you only send the the error logs to sentry in output/elastic module. I mean, you can send the application level error logs to sentry in goglog package which will imply for wider error logs.

Hi @tsaikd,
Changes have been done accordingly, Kindly review

cmd/module.go
```diff
@@ -41,7 +43,10 @@ func init() {
 		Use:   "worker",
 		Short: "gogstash worker mode",
 		RunE: func(ctx context.Context, cmd *cobra.Command, args []string) error {
-			return gogstash(ctx, flagConfig.String(), flagDebug.Bool(), flagPProf.String(), true)
+			err := gogstash(ctx, flagConfig.String(), flagDebug.Bool(), flagPProf.String(), true)
+			sentry.CurrentHub().Recover(err)
+			sentry.Flush(time.Second * 5)
```
tsaikd, line 48: Did you check the recover function work when the application panic? It's better to write some tests.

config/goglog/goglog.go
```diff
@@ -79,26 +81,31 @@ func (t LoggerType) Printf(format string, args ...any) {
...
 
 // Errorf wrap logrus function
 func (t LoggerType) Errorf(format string, args ...any) {
+	sentry.CaptureMessage(fmt.Sprintf(format, args...))
 	t.stderr.Errorf(format, args...)
 }
 
 // Fatalf wrap logrus function
 func (t LoggerType) Fatalf(format string, args ...any) {
+	sentry.CaptureMessage(fmt.Sprintf(format, args...))
```
tsaikd, line 102: Without log level?

cmd/module.go
cdmx1: We tested it, Since Sentry is an external dependency, writing unit tests to directly validate its behavior in case of panics would be challenging.

config/goglog/goglog.go
cdmx-lavi-sidana: Updated the log level for sentry

cdmx1:
@tsaikd , can you review it now?

tsaikd:
Simple tests to check sentry sdk usage.

ConfigureScope, WithScope do not work atomically. There are some messages with the wrong log level.


```go
func TestSentryParallel(t *testing.T) {
	err := sentry.Init(sentry.ClientOptions{
		Dsn:              "",
		TracesSampleRate: 1.0,
	})
	require.NoError(t, err)

	hubInfo := sentry.CurrentHub().Clone()
	hubInfo.ConfigureScope(func(scope *sentry.Scope) {
		scope.SetLevel(sentry.LevelInfo)
	})
	hubWarn := sentry.CurrentHub().Clone()
	hubWarn.ConfigureScope(func(scope *sentry.Scope) {
		scope.SetLevel(sentry.LevelWarning)
	})

	for i := 0; i < 60; i++ {
		i := i
		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
			t.Parallel()
			switch {
			case i%6 == 0:
				sentry.WithScope(func(scope *sentry.Scope) {
					scope.SetLevel(sentry.LevelInfo)
					sentry.CaptureMessage(fmt.Sprintf("Scope Info %d", i))
				})
			case i%6 == 1:
				sentry.WithScope(func(scope *sentry.Scope) {
					scope.SetLevel(sentry.LevelWarning)
					sentry.CaptureMessage(fmt.Sprintf("Scope Warn %d", i))
				})
			case i%6 == 2:
				sentry.ConfigureScope(func(scope *sentry.Scope) {
					scope.SetLevel(sentry.LevelInfo)
				})
				sentry.CaptureMessage(fmt.Sprintf("Configure Info %d", i))
			case i%6 == 3:
				sentry.ConfigureScope(func(scope *sentry.Scope) {
					scope.SetLevel(sentry.LevelWarning)
				})
				sentry.CaptureMessage(fmt.Sprintf("Configure Warn %d", i))
			case i%6 == 4:
				hubInfo.CaptureMessage(fmt.Sprintf("Hub Info %d", i))
			case i%6 == 5:
				hubWarn.CaptureMessage(fmt.Sprintf("Hub Warn %d", i))
			}
		})
	}

	t.Cleanup(func() {
		sentry.Flush(5 * time.Second)
		hubInfo.Flush(5 * time.Second)
		hubWarn.Flush(5 * time.Second)
	})
}
```

cmd/module.go
tsaikd: The test still panic.

ref: https://docs.sentry.io/platforms/go/usage/panics/

```go
func TestSentryRecover(t *testing.T) {
	err := sentry.Init(sentry.ClientOptions{
		Dsn:              "",
		TracesSampleRate: 1.0,
	})
	require.NoError(t, err)

	require.NotPanics(t, func() {
		err := func() error {
			panic("panic")
		}()
		sentry.CurrentHub().Recover(err)
		sentry.Flush(time.Second * 5)
	})
}
```

roney492:
@tsaikd , Tests updated. can you please review now?

tsaikd:
image

roney492:
@tsaikd , Fixes have been done, can you please review now?

config/goglog/goglog.go
```diff
@@ -79,26 +81,51 @@ func (t LoggerType) Printf(format string, args ...any) {
 
 // Warnf wrap logrus function
 func (t LoggerType) Warnf(format string, args ...any) {
+	hub := sentry.CurrentHub().Clone()
+	hub.ConfigureScope(func(scope *sentry.Scope) {
+		scope.SetLevel(sentry.LevelWarning)
+	})
+	hub.CaptureMessage(fmt.Sprintf(format, args...))
```
tsaikd, line 88: It's better to cache all hubs for the different levels.

cmd/module.go
```diff
@@ -40,8 +42,15 @@ func init() {
...
-		RunE: func(ctx context.Context, cmd *cobra.Command, args []string) error {
-			return gogstash(ctx, flagConfig.String(), flagDebug.Bool(), flagPProf.String(), true)
+		RunE: func(ctx context.Context, cmd *cobra.Command, args []string) (err error) {
+			defer func() {
+				if err != nil {
+					sentry.CurrentHub().Recover(err)
+					sentry.Flush(time.Second * 5)
+				}
+			}()
+			err = gogstash(ctx, flagConfig.String(), flagDebug.Bool(), flagPProf.String(), true)
```
tsaikd, line 52: It does not recover if `gogstash()` panic.

cdmx-lavi-sidana: As err given by gogstash function are handled  through module cobrather we can't recover it using sentry

roney492:
@tsaikd can you please review onces.

cmd/module.go
tsaikd: https://docs.sentry.io/platforms/go/usage/panics/

```go
func() {
	defer func() {
		err := recover()

		if err != nil {
			sentry.CurrentHub().Recover(err)
			sentry.Flush(time.Second * 5)
		}
	}()

	// do all of the scary things here
}()
```

config/goglog/goglog.go
```diff
@@ -38,13 +41,35 @@ func newLogger() *LoggerType {
...
+	stdout     *logrus.Logger
+	stderr     *logrus.Logger
+	sentryHubs map[sentry.Level]*sentry.Hub
+}
+
+// Helper function to get or create a Sentry hub for a specific level
+func (t *LoggerType) getSentryHub(level sentry.Level) *sentry.Hub {
+	if hub, ok := t.sentryHubs[level]; ok {
+		return hub
+	}
```
tsaikd, line 59: DATA RACE

```go
func TestLoggerParallel(t *testing.T) {
	err := sentry.Init(sentry.ClientOptions{
		Dsn:              "",
		TracesSampleRate: 1.0,
	})
	require.NoError(t, err)

	stdoutBuffer := &bytes.Buffer{}
	stderrBuffer := &bytes.Buffer{}
	logger := newLogger()

	logger.setDebugOutput(stdoutBuffer, stderrBuffer)
	logger.setDebugFormatter(&logrus.TextFormatter{DisableTimestamp: true})

	for i := 0; i < 5; i++ {
		i := i
		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
			t.Parallel()
			switch {
			case rand.Int()%2 == 0:
				logger.Warnf("Warn %d", i)
			case i%2 == 1:
				logger.Errorf("Error %d", i)
			}
		})
	}
}
```

cmd/module.go
cdmx-lavi-sidana: Upon reviewing the code, you will notice that we have implemented the same approach. The recover function proves useful, especially when applied to functions that may panic. During our tests on inputredis, we stopped the Redis server, resulting in an EOF error, which is the expected behavior. If we could somehow recover from this error, it would continuously ping Redis, consistently encountering the EOF. As by calling the gogstash function, which in turn calls additional functions it completes the whole execution once something returned and cobrather in doesn't panic it just simply prints the error and whole process is compleleted

tsaikd: Try to panic at cmd/gogstash.go:gogstash() to check if the recover function work as you want.

```diff

func gogstash(
	ctx context.Context,
	confpath string,
	debug bool,
	pprofAddress string,
	workerMode bool,
) error {
+	panic("A")
	if debug {
		goglog.Logger.SetLevel(logrus.DebugLevel)
	}
```

cdmx-lavi-sidana: What i meant it will recover if the code is implemented like below in functions

```
package main

import (
	"fmt"
	"log"
	"time"

	"github.com/getsentry/sentry-go"
)

func main() {
	err := sentry.Init(sentry.ClientOptions{
		Dsn: "YOUR_SENTRY_DSN", 
	})
	if err != nil {
		log.Fatalf("sentry.Init: %s", err)
	}
	defer sentry.Flush(2 * time.Second)

	go func() {
		defer recoverFromPanic()
		// Simulate a panic
		panic("something bad happened")
	}()

	// Continue running other parts of your application
	fmt.Println("Application continues running...")
	time.Sleep(10 * time.Second)
}

func recoverFromPanic() {
	if err := recover(); err != nil {
		sentry.CurrentHub().Recover(err)
		sentry.Flush(2 * time.Second)
		fmt.Println("Recovered from panic:", err)
	}
}
```

cdmx1: @tsaikd Is recovery really required? Shouldn't the application just exit on panic instead of continuing? If it's to continue, wouldn't retry be necessary?

config/goglog/goglog.go
cdmx-lavi-sidana: @tsaikd data race issue fixed

cmd/module.go
tsaikd: If you don't want to recover, why add this `sentry.CurrentHub().Recover(err)`?

config/goglog/goglog.go
```diff
@@ -38,13 +42,47 @@ func newLogger() *LoggerType {
...
+	stderr     *logrus.Logger
+	sentryHubs map[sentry.Level]*sentry.Hub
+	mu         sync.RWMutex
+}
+
+// Helper function to get or create a Sentry hub for a specific level
+func (t *LoggerType) getSentryHub(level sentry.Level) *sentry.Hub {
+	t.mu.RLock()
+	hub, ok := t.sentryHubs[level]
+	t.mu.RUnlock()
```
tsaikd, line 61: Mutex will add extra execution cost, and the log levels are fixed, so it's better to initialize all hubs first.

```go
type LoggerType struct {
	stdout        *logrus.Logger
	stderr        *logrus.Logger
	sentryHubInfo *sentry.Hub
	sentryHubWarn *sentry.Hub
	// ...
}
```

cmd/module.go
cdmx-lavi-sidana: @tsaikd we used recover here as it can decide either to capture Exception or Message, we have changed it to Capture Exception as its always going to be an error

config/goglog/goglog.go
cdmx-lavi-sidana: We have removed mutex, directly assiging hub to variable is not working for the second error so we have assigned function for Warn, fatal, Error

tsaikd:
```go

func newLogger() *LoggerType {
	warnHub := sentry.CurrentHub().Clone()
	warnHub.ConfigureScope(func(scope *sentry.Scope) {
		scope.SetLevel(sentry.LevelWarning)
	})
	// ...
}


// LoggerType wraps logrus.Logger type and adds a cache for Sentry hubs
type LoggerType struct {
	stdout         *logrus.Logger
	stderr         *logrus.Logger
	WarnSentryHub  *sentry.Hub
	// ...
}

func (t LoggerType) Warnf(format string, args ...any) {
	t.WarnSentryHub.CaptureMessage(fmt.Sprintf(format, args...))
	t.stdout.Warnf(format, args...)
}

func (t *LoggerType) Close() {
	sentry.CurrentHub().Flush(5 * time.Second)
	t.WarnSentryHub.Flush(5 * time.Second)
}
```

cmd/module.go
tsaikd: `defer` is not necessary if you do not recover.

```go
		RunE: func(ctx context.Context, cmd *cobra.Command, args []string) (err error) {
			err = gogstash(ctx, flagConfig.String(), flagDebug.Bool(), flagPProf.String(), false)
			if err != nil {
				hub := sentry.CurrentHub().Clone()
				hub.ConfigureScope(func(scope *sentry.Scope) {
					scope.SetLevel(sentry.LevelError)
				})
				hub.CaptureException(err)
				hub.Flush(5 * time.Second)
			}
			return err
		},
```

And it's better to move the logging codes to `gogstash()`.

config/goglog/goglog.go
roney492: @tsaikd can you review the changes now!

config/goglog/goglog.go
```diff
@@ -38,165 +55,206 @@ func newLogger() *LoggerType {
...
+		scope.SetLevel(level)
+	})
+	return hub
+}
+
+// LogWarning logs a warning message to Sentry
+func (l *LoggerType) LogWarning(msg string) {
+	l.WarnSentryHub.WithScope(func(scope *sentry.Scope) {
+		sentry.CaptureMessage(msg)
+	})
```
tsaikd, line 77:
```diff
-       l.WarnSentryHub.WithScope(func(scope *sentry.Scope) {
-               sentry.CaptureMessage(msg)
-       })
+       l.WarnSentryHub.CaptureMessage(msg)
```

config/goglog/goglog.go
```diff
@@ -38,165 +55,206 @@ func newLogger() *LoggerType {
...
+func (t *LoggerType) Printf(format string, args ...any) {
 	t.stdout.Printf(format, args...)
 }
 
 // Warnf wrap logrus function
-func (t LoggerType) Warnf(format string, args ...any) {
+func (t *LoggerType) Warnf(format string, args ...any) {
+	msg := fmt.Sprintf(format, args...)
+	t.WarnSentryHub.CaptureMessage(msg)
+	t.WarnSentryHub.Flush(time.Second * 3)
```
tsaikd, line 129: Flush every log?

config/goglog/goglog.go
```diff
@@ -38,165 +55,206 @@ func newLogger() *LoggerType {
...
 
 // Printf wrap logrus function
-func (t LoggerType) Printf(format string, args ...any) {
+func (t *LoggerType) Printf(format string, args ...any) {
 	t.stdout.Printf(format, args...)
 }
 
 // Warnf wrap logrus function
-func (t LoggerType) Warnf(format string, args ...any) {
+func (t *LoggerType) Warnf(format string, args ...any) {
```
tsaikd, line 126: Why pointer?

config/goglog/goglog.go
```diff
@@ -38,165 +55,206 @@ func newLogger() *LoggerType {
...
 	if err != nil {
 		t.Error(err)
-
 		return true
 	}
-
 	return false
 }
 
-var _ logrus.FieldLogger = &LoggerType{}
```
tsaikd, line 196: Why removing the type checker?

config/goglog/goglog.go
roney492: LoggerType is a large struct, copying it on each method call can be expensive. Using a pointer receiver avoids this overhead.

config/goglog/goglog.go
roney492: its not removed, its moved up.

cdmx-lavi-sidana:
@tsaikd Could you review it now, We have removed flush after each call , And You might again point out data race issue as we implemented the same code, Reason behind that is we tried assigning hubs for Error, Warn and fatal in struct itself, but problem occurs as goglog is initialised before the Sentry  and its assigned to variable now we are left with two options either we create new functions for all hubs which we need or the dynamic one as we created earlier and the other choice which we have already implemented this way its assigned once to struct it can be reused. If there is any other please let us know we will definetly take that approach

cdmx-lavi-sidana:
Hi @tsaikd, we've made the changes you suggested by removing flush after each call. Regarding the data race issue previously raised, we couldn't find a better solution. We tried assigning hubs for Error, Warn, and Fatal directly in the struct, but faced issues because goglog initializes before Sentry. Our current approach assigns these hubs once to the struct for reuse, or alternatively, we can create new functions for each hub. If you have any other suggestions, please let us know. We're open to exploring different approaches.

tsaikd:
What's your use case? Do you want to log into sentry from some inputs (e.g. files/docker), or log the gogstash app execution messages into sentry?

cdmx1:
> What's your use case? Do you want to log into sentry from some inputs (e.g. files/docker), or log the gogstash app execution messages into sentry?

Often, we encounter failures or issues with ingestions to OpenSearch/Elasticsearch without us even noticing, in such cases sentry will keep us informed of us failures, specially wherein we are using Docker and gogstash container recovers from the failures on docker restart polices, This will help us stay aware of any issues, allowing us to take appropriate actions, such as resolving field mapping or index conflicts.

tsaikd:
> we are using Docker and gogstash container

Do you mean you are running the gogstash in a docker container to collect the other application logs (e.g. nginx) and save to the ElasticSearch?
What's your input module in the gogstash?

cdmx1:
> > we are using Docker and gogstash container
> 
> Do you mean you are running the gogstash in a docker container to collect the other application logs (e.g. nginx) and save to the ElasticSearch? What's your input module in the gogstash?

Yes, we use a Docker container for gogstash to collect logs from various sources like Node.js applications and Nginx. These logs are initially ingested into Redis, and then fed into gogstash using Redis as an input. Gogstash handles the flushing of logs to OpenSearch/Elasticsearch. However, we've observed frequent failures on the OpenSearch/Elasticsearch side due to incorrect data types sent from the applications. These failures often go unnoticed.

tsaikd:
> We tried assigning hubs for Error, Warn, and Fatal directly in the struct, but faced issues because goglog initializes before Sentry.

In your use case, the sentry should not be an output module, and that's why you have the issue.

You can put the sentry config at `config/config.go:Config` as the gogstash application level config, initialize global sentry at `config/config.go:initConfig()`, save sentry hub instances at `config/goglog/goglog.go:LoggerType`, and flush at `config/config.go:Config.Wait()`

cdmx1:
> > We tried assigning hubs for Error, Warn, and Fatal directly in the struct, but faced issues because goglog initializes before Sentry.
> 
> In your use case, the sentry should not be an output module, and that's why you have the issue.
> 
> You can put the sentry config at `config/config.go:Config` as the gogstash application level config, initialize global sentry at `config/config.go:initConfig()`, save sentry hub instances at `config/goglog/goglog.go:LoggerType`, and flush at `config/config.go:Config.Wait()`

Hi @tsaikd,
Changes made, can you please review

cmd/module.go
tsaikd: @cdmx-lavi-sidana

cdmx-lavi-sidana:
@tsaikd As suggested, we tried initialising the sentry in config, but the actual problem was with goglog, so we have changed how goglog will be loaded initially, now the variable for different sentry hub is working fine, but to fix this we had to nove sentry to gogstash function. Could you please review now

cmd/module.go
tsaikd: > `defer` is not necessary if you do not recover.
> 
> ```go
> 		RunE: func(ctx context.Context, cmd *cobra.Command, args []string) (err error) {
> 			err = gogstash(ctx, flagConfig.String(), flagDebug.Bool(), flagPProf.String(), false)
> 			if err != nil {
> 				hub := sentry.CurrentHub().Clone()
> 				hub.ConfigureScope(func(scope *sentry.Scope) {
> 					scope.SetLevel(sentry.LevelError)
> 				})
> 				hub.CaptureException(err)
> 				hub.Flush(5 * time.Second)
> 			}
> 			return err
> 		},
> ```
> 
> And it's better to move the logging codes to `gogstash()`.

@cdmx-lavi-sidana do you see the comment?

cdmx-lavi-sidana: @tsaikd its updated, Please check

PR #241, opened 2025-03-03
Feature/output http options

jbuchbinder:
Add ability to set custom headers and content-type.

Might be a way dirtier patch than necessary.


tsaikd:
Please remove the unnecessary commits for the MR.

PR #244, opened 2025-03-27
[codec] prometheus : initial implementation of prometheus wire codec

jbuchbinder:


tsaikd:
Please fix lint issues:

https://github.com/tsaikd/gogstash/actions/runs/14107953286/job/39518884619?pr=244

jbuchbinder:
Lint issues fixed, as requested.

codec/prometheus/codecprometheus.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+
+		// Decode lines, every metric is an possible individual event
+		{
+			e := event
+			err = c.decodePrometheusEvent(line, savetype, &e)
+			if err == nil {
+				msgChan <- e
+				ok = true
+			}
```
tsaikd, line 104: ignore decodePrometheusEvent error?

jbuchbinder: That only pushes events into the channel if there are no errors in decoding.

codec/prometheus/codecprometheus.go
```diff
@@ -0,0 +1,225 @@
...
+			continue
+		}
+
+		// Decode lines, every metric is an possible individual event
+		{
+			e := event
+			err = c.decodePrometheusEvent(line, savetype, &e)
+			if err == nil {
+				msgChan <- e
+				ok = true
```
tsaikd, line 104:
```suggestion
				select {
				case <-ctx.Done():
					// error handling ...
				case msgChan <- e:
					ok = true
				}
```

codec/prometheus/codecprometheus.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+		// Skip line if comment, we don't process them further
+		if strings.HasPrefix(line, "#") {
+			continue
+		}
+
+		// Decode lines, every metric is an possible individual event
+		{
+			e := event
+			err = c.decodePrometheusEvent(line, savetype, &e)
```
tsaikd, line 101:
```go
	e1 := logevent.LogEvent{
		Extra: map[string]any{
			"foo": "bar",
		},
	}
	e2 := e1
	e2.Extra["foo"] = "omg" // e1.Extra["foo"] is also assigned to "omg"
```

Is it what you want?

jbuchbinder: It's supposed to be a copy, not a reference. Rather than iteratively instantiating logevent.LogEvent{} objects, I was just trying to copy them and then modify then with the actual events.

codec/prometheus/codecprometheus.go
```diff
@@ -0,0 +1,229 @@
...
+			if err == nil {
+				select {
+				case <-ctx.Done():
+					// error handling ...
```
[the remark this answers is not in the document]
jbuchbinder: Probably not, since it would indicate the connection being closed.

PR #250, opened 2025-11-27
feat(output-clickhouse): add native ClickHouse output plugin with TLS…

fabiodepin:
feat(output-clickhouse): add native ClickHouse output plugin with TLS support

This commit introduces a new output module  for gogstash, providing native integration with ClickHouse using HTTP/HTTPS and the JSONEachRow format.

Key features:
- Support for multiple ClickHouse URLs (load-balanced per flush)
- Optional Basic Authentication
- Configurable batch size and flush interval
- Optional timestamp mapping via
- Optional raw event preservation
- Optional strict error propagation ()
- TLS option:  (allow self-signed certificates).

### DIFF
--- a/output/clickhouse/outputclickhouse.go
+++ b/output/clickhouse/outputclickhouse.go
@@ -1,0 +1,310 @@
+package outputclickhouse
+
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"encoding/base64"
+	"fmt"
+	"io"
+	"math/rand"
+	"net/http"
+	"sync"
+	"time"
+
+	jsoniter "github.com/json-iterator/go"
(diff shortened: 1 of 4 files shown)


output/clickhouse/outputclickhouse.go
```diff
@@ -0,0 +1,310 @@
...
+// flushLoop performs time-based flushing
+func (t *OutputConfig) flushLoop(ctx context.Context) {
+	ticker := time.NewTicker(t.flushInterval / 2)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			// Final flush on shutdown
+			_ = t.flush(ctx)
```
tsaikd, line 196: at the moment, the ctx is done, so the flush(ctx) may not work as you wish.

